PR #159, opened 2015-08-10
Added FCOPY_non_ascii testcase

adriansuhov:
FCOPY_non_ascii  will generate a 100MB file with non-ascii characters.
Then it will copy the file to the Linux VM. Finally, the script will
verify both MD5 checksums (on host and guest).

### DIFF
--- a/WS2012R2/lisa/setupscripts/FCOPY_non_ascii.ps1
+++ b/WS2012R2/lisa/setupscripts/FCOPY_non_ascii.ps1
@@ -1,0 +1,257 @@
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved. 
+# Licensed under the Apache License, Version 2.0 (the ""License"");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0  
+#
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+# OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/setupscripts/FCOPY_non_ascii.ps1
```diff
@@ -0,0 +1,257 @@
...
+    "Error: The Guest services are not working properly for VM '${vmName}'!" | Tee-Object -Append -file $summaryLog
+	$retVal = $False
+}
+else {
+	# Define the file-name to use with the current time-stamp
+    $MyDir= "$pwd\"
+	$testfile = "testfile-$(get-date -uformat '%H-%M-%S-%Y-%m-%d').file" 
+    $pathToFile="$MyDir"+"$testfile" 
+
+	# Create a ~100MB sample file with non-ascii characters
```
chvalean: File should be generated only after it's confirmed that the fcopy daemon is present/running first.

WS2012R2/lisa/setupscripts/FCOPY_non_ascii.ps1
```diff
@@ -0,0 +1,257 @@
...
+    $stream = [System.IO.StreamWriter] $pathToFile
+    $s="¡¢£¤¥§¨©ª«¬®¯±µ¶←↑ψχφυ"
+    1..1000000 | % {
```
chvalean: the test file only gets to a size of 38M, is there a buffer limitation somewhere?

WS2012R2/lisa/setupscripts/FCOPY_non_ascii.ps1
```diff
@@ -0,0 +1,257 @@
...
+
+#######################################################################
+#
+#	Checks if test file is present
+#
+#######################################################################
+function check_file([String] $testfile)
+{
+    $localChksum = Get-FileHash .\$testfile -Algorithm MD5 | Out-String
+    $localChksum = $localChksum.Substring(450,32)
```
chvalean: function is incorrect, must only filter the hash property on line 102

PR #160, opened 2015-08-13
VHD/X differencing disk automation

thenoizz:
Automating the ParentVHD creation, if it's not supplied in the testparams. Depending on the `vhdFormat` parameter, the script creates either a vhdx or a vhd parent disk.


WS2012R2/lisa/setupscripts/DiffDiskGrowthTestCase.ps1
```diff
@@ -138,7 +297,7 @@ if ($hvModule -eq $NULL)
 if ($hvModule.companyName -ne "Microsoft Corporation")
```
chvalean: No need to perform this check

thenoizz: ok.

PR #178, opened 2015-09-04
Use regular expressions to parse <memSize>. Set min memory for new VM to 512MB. Cleanup extra blank lines. Add initial checks to support the <ImportVM> attribute.

nickme:


WS2012R2/lisa/setupscripts/CreateVMs.ps1
```diff
@@ -320,90 +334,77 @@ function CheckRequiredParameters([System.Xml.XmlElement] $vm)
...
+    #        "2048MB"       "2048GB"
+    #        "2040 MB"      "2048 GB"
+    #
+    if ($vm.hardware.memSize)
+    {
+        #
+        #    Use regular expressions to parse the memory size string
+        #    and convert the value to MB.  Whitespace is parsed out.
+        #
+        $regex = [regex] '^(\d+)\s*([MG]B)?$'
```
alexngmsft: Would be nice to handle trailing whitespaces, but this is fine.

PR #196, opened 2015-10-23
Fix static MAC address setting and add support for copying parent VHD…

alexngmsft:
… to remote Hyper-V server.


chvalean:
Will merge as part of Zsolt commit, to not break that commit.

PR #239, opened 2016-01-26
Add the branch of RedHat version

Jingli1985:
1. There are different versions in Redhat  for example: RHEL6 and RHEL7. however, different commands are suitable for different versions.  so split them in file. 
2. the "Red*Hat*" can't match the result.

### DIFF
--- a/WS2012R2/lisa/setupscripts/ProvisionSshKeys.ps1
+++ b/WS2012R2/lisa/setupscripts/ProvisionSshKeys.ps1
@@ -119,7 +119,7 @@ function GetLinuxDistro([String] $ipv4, [String] $password)
         return $null
     }
 
-    $distro = bin\plink -pw "${password}" root@${ipv4} "grep -hs 'Ubuntu\|SUSE\|Fedora\|Debian\|CentOS\|Red Hat Enterprise Linux\|Oracle' /etc/{issue,*release,*version}"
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/setupscripts/ProvisionSshKeys.ps1
```diff
@@ -170,7 +173,8 @@ function InstallPackagesRequiredByLisa([String] $ipv4, [String] $password)
     switch -regex ($distro)
     {
         CentOS   { $cmd = "yum -y install dos2unix at"                        }
-        RedHat   { $cmd = "yum -y install dos2unix at"                        }
+        RedHat6   { $cmd = "yum -y install dos2unix at"                        }
+        RedHat7   { $cmd = "yum -y install dos2unix at"                        }
```
chvalean: We should keep the general RedHat tag for all versions from 5.x to 7.x. This is also to keep the compatibility with the current scripts that use the RedHat distro regex.
A 3rd reason is to remove duplicate code like the same commands for RHEL 6.x and 7.x if nothing is changed.

Jingli1985: You are right, I will remove that.

chvalean:
Script might be updated to use GetOSVersion from utils.sh, which uses a more consistent distro identifier.

PR #256, opened 2016-02-29
Added NET_HotAddRemove testcase

adriansuhov:
1. Split up NET_HotAddMultiNIC in 2 testcases: First one checks the NIC
Hot Add capability and second one checks for NIC Hot Remove capability.
2. Added RHEL and Ubuntu support for both testcases

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/NET_VerifyHotAddMultiNIC.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/NET_VerifyHotAddMultiNIC.sh
@@ -23,6 +23,99 @@
 
 
 expectedCount=0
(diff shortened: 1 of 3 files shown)


WS2012R2/lisa/remote-scripts/ica/NET_VerifyHotAddMultiNIC.sh
```diff
@@ -23,6 +23,99 @@
 
 
 expectedCount=0
+declare os_VENDOR os_RELEASE os_UPDATE os_PACKAGE os_CODENAME
+
+########################################################################
+# Determine what OS is running
+########################################################################
+# GetOSVersion
```
chvalean: Please source this function from utils.sh - include this in the xml definition and source it accordingly in the script.

PR #273, opened 2016-03-16
Add verifying VCPU can't be offline testcase

Jingli1985:
Add case to make sure the vcpu can't be offline on Guest. The Hyper-V 2012 R2 don't support CPU onlining or offlining, so the guest also can't be offline.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/vcpu_verify_online.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/vcpu_verify_online.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved. 
+# Licensed under the Apache License, Version 2.0 (the ""License"");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0  
+#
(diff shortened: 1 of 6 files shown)


WS2012R2/lisa/remote-scripts/ica/vcpu_verify_online.sh
```diff
@@ -0,0 +1,130 @@
...
+    source ${CONSTANTS_FILE}
+else
+    msg="Error: no ${CONSTANTS_FILE} file"
+    LogMsg $msg
+    echo $msg >> ~/summary.log
+    UpdateTestState $ICA_TESTABORTED
+    exit 10
+fi
+
+if [ -e ~/summary.log ]; then
```
chvalean: this needs to be moved before sourcing constants, as there we write to summary already

WS2012R2/lisa/remote-scripts/ica/vcpu_verify_online.sh
```diff
@@ -0,0 +1,130 @@
...
+echo "This script covers test case: ${TC_COVERED}" >> ~/summary.log
+
+#
+# Getting the CPUs count
+#
+cpu_count=$(grep -i processor -o /proc/cpuinfo | wc -l)
+echo "${cpu_count} CPU cores detected" >> ~/summary.log
+
+#
+# Verifying all CPUs can't be online except CPU0
```
chvalean: can't be offline I presume, minor typo

chvalean:
Thanks Jing for the test case. After you get the chance to verify my comments to be files changed, I can merge the changes.

PR #277, opened 2016-03-18
Add Redhat kernel upgrade and downgrade testcase

Jingli1985:


WS2012R2/lisa/xml/RedHat_test_kernel.xml
```diff
@@ -0,0 +1,73 @@
...
+    </testSuites>
+
+    <testCases>
+        <test>
+            <testName>redhat_update_kernel</testName>
+            <testScript>setupscripts\kernel.ps1</testScript>
+            <files>remote-scripts\ica\redhat_upgrade_kernel.sh,remote-scripts\ica\redhat_downgrade_kernel.sh,remote-scripts\ica\utils.sh</files>
+            <testParams>
+                <!--<param>URL=http://PATH/TO/TEST/KERNEL</param>-->
+                <param>URL=http://10.66.10.35/pub/kernel/</param>
```
chvalean: is this a private/internal URL?

WS2012R2/lisa/xml/RedHat_test_kernel.xml
```diff
@@ -0,0 +1,73 @@
...
+                <param>DOWNKERNEL=kernel-3.10.0-305.el7.x86_64.rpm</param>
+                <param>UPKERNEL=kernel-abi-whitelists-3.10.0-364.el7.noarch.rpm,kernel-3.10.0-364.el7.x86_64.rpm</param>
```
chvalean: my concern here is that the script is specifically targeting 7.2 only.
if this can be made more generic to run at least on any 7.x or even some of the latest 6.x (6.7 maybe), it would make the test more broadly to be used.

WS2012R2/lisa/remote-scripts/ica/redhat_downgrade_kernel.sh
```diff
@@ -0,0 +1,138 @@
...
+
+#
+# Make sure constants.sh contains the variables we expect
+#
+if [ "${TC_COVERED:-UNDEFINED}" = "UNDEFINED" ]; then
+    msg="The test parameter TC_COVERED is not defined in ${CONSTANTS_FILE}"
+    LogMsg $msg
+    echo $msg >> ~/summary.log
+    UpdateTestState $ICA_TESTABORTED
+    exit 30
```
chvalean: not having the TC_COVERED param is not a test breaker, so a warning only should be enough

WS2012R2/lisa/remote-scripts/ica/redhat_downgrade_kernel.sh
```diff
@@ -0,0 +1,138 @@
...
+        wget $URL/$file
+        rpm -Uvh $file --oldpackage
+	    if [[ $? -ne 0 ]]; then
+		    UpdateSummary "Error: Unable to install the test kernel ${file}!"
+		    UpdateTestState $ICA_TESTABORTED
+		    exit 1
+	    fi
+	done
+
+	echo "Info: The kernel dependecies pacakages have been successfully installed!" >> ~/summary.log
```
chvalean: There should also be a grub change to ensure that for either the upgrade or downgrade kernel scenarios, the targeted kernel at reboot is to be set as the one to boot.
We use this logic for the ubuntu test and proposed kernels, where we also upgrade grub to boot a specific kernel.

Jingli1985: 1. I installed kernel with parameter "-U", there is only one menu list in grub, so I didn't do the check in shell script.
2. In powershell script , To check if the kernel version is right or not with code below:
`227 $kernelPackage =  echo y | bin\plink -i ssh\${sshKey} root@${ipv4} "uname -r"
228 if (not $kernelDownVersion -match $kernelPackage)
229 {
230     Write-Output "The Kernel version have some problem"
231     return $False
`

WS2012R2/lisa/xml/RedHat_test_kernel.xml
Jingli1985: The script isn't specifically targeting 7.2 only, we can set any version of kernel need to upgrade or downgrade for VM(rhel7.2 to rhel7.3) , we also can do upgrade or downgrade for different kernel in rhel7.2.

WS2012R2/lisa/xml/RedHat_test_kernel.xml
Jingli1985: sorry, I will modify it

PR #301, opened 2016-04-14
Add case of numa can work when vCPU or Memory exceed the NUMA node ma…

Jingli1985:
1. Add a new case in xml "Verify numa can work when vCPU or Memory exceed the NUMA node maximum."
2. fix a issue to get the maximums of supported vCPUs per host. 
3. fix to calculate the expected_number of node.

### DIFF
--- a/WS2012R2/lisa/setupscripts/ChangeCPU.ps1
+++ b/WS2012R2/lisa/setupscripts/ChangeCPU.ps1
@@ -40,7 +40,47 @@
 #>
 
 param([string] $vmName, [string] $hvServer, [string] $testParams)
+########################################################################
+#
+# ConvertStringToDecimal()
+#
+########################################################################
+function ConvertStringToDecimal([string] $str)
+{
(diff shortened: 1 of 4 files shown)


WS2012R2/lisa/setupscripts/ChangeCPU.ps1
```diff
@@ -105,11 +151,11 @@ if ($procs)
 {
     if ($procs -is [array])
     {
-        $maxCPUs = $procs[0].NumberOfCores
+        $maxCPUs = $procs[0].NumberOfLogicalProcessors *2
```
chvalean: Switching to NumberOfLogicalProcessors makes sense, but why is it multiplied by 2? The physical CPUs x2 give the logical cores count.

Jingli1985: you are right. the NumberOfLogicalProcessor shouldn't be multiplied by 2.

PR #311, opened 2016-04-29
Call Traces for DM & Modules check in initrd image

OvidiuRusu:
Check for Call Traces in DM and Modules check in initrd image based on the missing keyboard module bug


nickme:
The new file 'Check_traces.sh' needs the license comments included at the top of the file.

PR #317, opened 2016-05-10
1. Fixed logMsg variable error, a ')' missed; 2. Fixed for loop error…

Jingli1985:
1. Fixed logMsg variable error, a ')' missed; 
2. Fixed for loop error, which could not fetch the last element in array.


chvalean:
Thank you Jing for the additions!

PR #319, opened 2016-05-10
Added new test case, which verifies the NUMA when kernel paramter is off

Jingli1985:


WS2012R2/lisa/remote-scripts/ica/NUMA_off_kernel.sh
```diff
@@ -0,0 +1,158 @@
...
+# Source constants file and initialize most common variables
+UtilsInit
+
+# Get distro
+GetDistro
+
+#
+# Add "numa=off" into kernel boot parameter
+#
+case $DISTRO in
```
chvalean: We try to design test cases to handle different distributions, unless it is really a distro specific feature.
Would you be able to add a distro case at least for ubuntu and sles as well?
Otherwise we can make that addition, however we prefer to cover at least these 3 main distributions.
Thank you!

Jingli1985: Hi Chris,
We would like take more energy on Rhel, for other distribution I don't have deeply acknowledge. it is better to add by owner I think
Thanks

nickme:
There are conflicts.  Would you resolve the conflicts and we will merge the request.

PR #320, opened 2016-05-10
Add VHDx_IDE_SCSI_Mixed_Dynamic case

Jingli1985:
Add a case VHDx_IDE_SCSI_Mixed_Dynamic in STOR_VHDX to cover scsi and IDE mixed


chvalean:
The test case does cover a scenario not covered directly, however as it's applicable to generation 1 VMs, it will always be covered as the system disk resides on IDE 0-0.
So testing this with IDE 0-1 doesn't necessarily bring a real case not covered.

PR #322, opened 2016-05-12
fix the list of checking modules

Jingli1985:
The hv_utils.ko that used to connect with integration serviceisn't necessary in VM temporary booting process, however the hyperv_fb.ko should add into checking list, so that the it is used when VM is booted.

### DIFF
--- a/WS2012R2/lisa/xml/CoreTests.xml
+++ b/WS2012R2/lisa/xml/CoreTests.xml
@@ -289,7 +289,7 @@
             <files>remote-scripts/ica/modules_check.sh</files>
             <testparams>
                 <param>TC_COVERED=CORE-24</param>
-                <param>hv_modules=(hv_vmbus.ko hv_storvsc.ko hv_netvsc.ko hyperv-keyboard.ko hid-hyperv.ko hyperv_fb.ko)</param>
+                <param>hv_modules=(hv_vmbus.ko hv_storvsc.ko hv_netvsc.ko hyperv-keyboard.ko)</param>
             </testparams>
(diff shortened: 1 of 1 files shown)


chvalean:
This test originated from the need of having the hyperv keyboard driver, as that would block any input if not loaded. The rest of the modules are part of sanity check, so my proposal is to keep both the utils and hyperv fb driver.
Thank you Jing!

nickme:
To boot a VM in a typical configuration, you need hv_vmbus.ko and hv_storvsc.ko.  If the root partition is encrypted, you will need hyperv_keyboard.ko so you can enter the dm_crypt passphrase.  So these three drivers are the minimum set that should be in the initrd.  If your VM is configured as a diskless client, you will need hv_netvsc.ko in the initrd as well.  So the minimum set of drivers the test case should check for would be:
    hv_vmbus.ko
    hv_storvsc.ko
    hyperv_keyboard.ko
    hv_netvsc.ko
Additional LIS drivers can be included in the initrd, but I do not see a need for the test case to check for them.  The remaining LIS drivers can be loaded from the root partition once it is mounted.

Jingli1985:
Hi Nick, 
Here is a list in rhel7 . All mandatory modules are be included in installation initramfs:
hv_vmbus.ko
hv_netvsc.ko
hv_storvsc.ko
hid-hyperv.ko
hyperv-keyboard.ko
hyperv_fb.ko
The hyperv_fb.ko is also included in list in order to the driver is used when the VM is booted. It is ok to check  the minimum set for this case, also is applicable for all of distribution.

PR #330, opened 2016-05-25
Added a new case, which verifies whether the NUMA is turned off 

Jingli1985:
"'numa=off' was added as kernel boot params.


Jingli1985:
Hi Nick, 
I closed the previous pull request and reopen new one, the conflict has fixed

Thanks

PR #339, opened 2016-06-06
Provision client vm

Jingli1985:
1. Some cases during testing process need to use some storage like ISCSI and NFS, Add a script of deploy vm Client to set up NFS and ISCSI server automatically meanwhile also add an Internal Switch for the related Network performance cases. 
2 Add default location support if it is not set . 
3. Fixed stop firewall issue for RHEL7
4. Some cases refer to Network need to set the 2rd VM(IP), For some cases need to deploy env , we can do it ahead in provisioning client VM


Jingli1985:
Hi Chris, 
I have fixed it and pushed current branch, Could you merge it or do I need reopen a new pull request?

chvalean:
I don't see the push, so probably a new PR would be needed.

Jingli1985:
@chvalean  have pushed #351 PR, thanks

PR #346, opened 2016-06-12
STOR_VHDX: Add two cases about copy 5G files( copy locally/wget/nfs) …

Jingli1985:
…to SCSI/IDE disk


chvalean:
Hi Jing Li,
Thank you for the new additions!
We would like to take these scenarios, but have a few recommendations first:
- since the initial commit some other changes seems to have occurred, and we cannot merge it in the current state.
- can the test be modified to handle selective tests - for eg run only wget?
- there should be no effect in testing both IDE and SCSI for these scenarios, so my recommendation is to focus only on SCSI, so one test case.

lixuemin2016:
Hi Chris,
Thank you so much for your review, will update new patch based on your recommendation.

Jingli1985:
Close this PR, will request a new one

PR #349, opened 2016-06-12
Storage logical sector

Jingli1985:
add one case about checking the logical and physical sector size for disk with 512 sector size


chvalean:
@Jingli1985 please re-submit this one based on the new storage tests params and I'll accept this PR.
Thank you!

PR #350, opened 2016-06-13
Add case of check and uncheck ExchangeService

Jingli1985:
Disable and Enable Key-Value Pair Exchange serice and Verify the basic KVP read operations work
Here is a bug1074407 on less RHEL7.2 . This case must be run on more than RHEL7.2

### DIFF
--- a/WS2012R2/lisa/setupscripts/INST_LIS_ExchangeServiceDisableEnable.ps1
+++ b/WS2012R2/lisa/setupscripts/INST_LIS_ExchangeServiceDisableEnable.ps1
@@ -1,0 +1,286 @@
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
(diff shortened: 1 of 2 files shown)


chvalean:
Hi @Jingli1985,
This seems to be a variation of the basic KVP check, with the Data Exch service actions.
In order to avoid too much duplicate code, my request would be if you'll be able to integrate the DE disable/enable in a function and have that ran based on the new test case param like DE_change=yes.
This would then just re-use the kvp basic script, with this addition only.
Please let me know if you can accommodate the changes or otherwise we can do it from our end.
Thank you!

Jingli1985:
Hi @chvalean 
I can do accommodate this change. I will do it later, I think the case of uncheck and check integration to check timekeeping also can do similar change, Do you think so?

Jingli1985:
Close this PR, As Chris advice to solve this in another PR.

PR #352, opened 2016-06-13
remove unused paramater

Jingli1985:


WS2012R2/lisa/setupscripts/NET_ADD_Switch.ps1
```diff
@@ -45,38 +45,12 @@
    All setup and cleanup scripts must return a boolean ($true or $false)
    to indicate if the script completed successfully or not.
 
-   .Parameter vmName
-	Name of the VM to add NIC to .
-
-	.Parameter hvServer
```
chvalean: Please let's keep the hvServer param and any code should be tested on a remote hvServer.
This is to cover cases with the nano, server core or Hyper-V Server.

WS2012R2/lisa/setupscripts/NET_ADD_Switch.ps1
```diff
@@ -109,7 +83,7 @@ foreach ($p in $params)
         $switchName = $switchArgs[1].Trim()
 
         #
-        # Validate the virtual switch type
+        # Validate the virutal switch type
```
chvalean: typo here and below, these have been correct after the initial push

PR #355, opened 2016-06-14
Add case of checking timekeeping after multiple reboot

Jingli1985:


chvalean:
Thanks Jing, the test case does seem relevant in what it does.
For the implementation I find it too be a little bit more long and prone to errors between distros.
What I would recommend is to design it around the setupscripts/Test50TimesReboot.ps1 script.
That one triggers the reboot operations from the host side, so it's distro agnostic.
Plus that the code is already there and can be merely only modified as to run maybe even as a setup or pretest script, then have the actual time sync check script.
Thank you!

Jingli1985:
Thanks your advice, Chris, I didn't notice this script . It is so convenient to use exist script I just need to modify xml to add this case.

Jingli1985:
Close this PR to add case with exist script, Thank Chris

PR #357, opened 2016-06-15
STOR_VHDXResize: add three cases about resize IDE disk

Jingli1985:


WS2012R2/lisa/setupscripts/STOR_VHDXResize_Basic.ps1
```diff
@@ -68,6 +70,8 @@ $TC_COVERED = $null
 $TestLogDir = $null
 $TestName   = $null
 $vhdxDrive  = $null
+$controllerType = "SCSI"
```
chvalean: this should not be hardcoded here, as the value from xml will be then ignored.

WS2012R2/lisa/setupscripts/STOR_VHDXResize_Basic.ps1
```diff
@@ -154,12 +160,13 @@ Write-Output "Covers: ${TC_COVERED}" | Tee-Object -Append -file $summaryLog
 # Convert the new size
 #
 $newVhdxSize = ConvertStringToUInt64 $newSize
-
+$sizeFlag = ConvertStringToUInt64 "50GB"
```
chvalean: what is the role of the new var sizeFlag? The 'newvhdxsize' variable is currently used and parsed from the xml

WS2012R2/lisa/setupscripts/STOR_VHDXResize_Basic.ps1
lixuemin2016: Thank you so much for your comment, I will update it by removing controllerType hardcode here and adding value to xml for all related cases .

WS2012R2/lisa/setupscripts/STOR_VHDXResize_Basic.ps1
lixuemin2016: Thank you so much for your comment, here mainly in order to support large disk size, e.g. 2T, and refer to file STOR_VHDXResize_DiskOver2TB.ps1, merge the resize disk function to one file.
1) If host  does not have enough space e.g. >2T, use sizeFlag "50GB" to check free space even it is not so accurate, but can indicate the machine has very less free space, if directly uses newVhdxSize,  test case may abort directly.
2) For dynamic disk, it needs less free space .
+if ($diskInfo.FreeSpace -le $sizeFlag + 10MB)
  {
      "Error: Insufficent disk free space"
      "  This test case requires ${newSize} free" ....
      return $False
 }

PR #373, opened 2016-07-05
fix mirror

Jingli1985:


Jingli1985:
Hi @chvalean 
There is an error when covert $line to ASCII, so recover here. but if we won't covert it. there are some can't be identified character when the log is stored in linux

PR #380, opened 2016-07-06
Add a new state - ApplyCheckpoint, which restore snapshot if configur…

Jingli1985:
ApplyCheckpoint state is after SystemDown and before RunSetupScript or StartSystem.

### DIFF
--- a/WS2012R2/lisa/stateEngine.ps1
+++ b/WS2012R2/lisa/stateEngine.ps1
@@ -197,6 +197,7 @@ foreach($MN in $MNs)
 # States a VM can be in
 #
 New-Variable SystemDown          -value "SystemDown"          -option ReadOnly
+New-Variable ApplyCheckpoint     -value "ApplyCheckpoint"     -option ReadOnly
 New-variable RunSetupScript      -value "RunSetupScript"      -option ReadOnly
 New-Variable StartSystem         -value "StartSystem"         -option ReadOnly
 New-Variable SystemStarting      -value "SystemStarting"      -option ReadOnly
@@ -559,6 +560,12 @@ function DoStateMachine([XML] $xmlConfig)
                     $done = $false
                 }
 
+            $ApplyCheckpoint
+                {
+                    DoApplyCheckpoint $vm $xmlConfig
+                    $done = $false
+                }
+
             $RunSetupScript
                 {
                     DoRunSetupScript $vm $xmlConfig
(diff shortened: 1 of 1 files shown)


chvalean:
What is the use case for this new state?
As for restoring snapshots, we make use of a setup script definition, like for the example below.
Is this model not usable in your use cases?

```
<setupScript>
                <file>setupscripts\RevertSnapshot.ps1</file>
                <file>setupscripts\NET_ADD_NIC_MAC.ps1</file>
</setupScript>
```

I can merge this change, but we want first to understand the full benefit of it and ensure this doesn't add something that is already there.
Thank you!

Jingli1985:
Hi Chris, 
The function of this patch is same as your example above, we just do it with another ways. the main framework do this instead of in setupScript process. for example, if every case need to do restore before running testing that need to add script with <file>setupscripts\RevertSnapshot.ps1</file>, we think it is a little confusion， too many same codes is in setupscript. however, Do this in main flow instead, just need to add it with param "noCheckpoint=False" in every case, Also it is easy to add illustration in document.

alexngmsft:
I agree with the general idea of having “Revert” be part of the XML language (to improve readability and usability).

I would suggest the following changes to JingLi’s code:

- Rename this option to explicitly state what it does and avoid confusion. A clearer name might be “RevertDefaultSnapshot”.
- Have the flag be opt-in to avoid changing existing test case behavior. If it exists and is set to “true”, then revert the snapshot. Otherwise, don’t do anything.

WS2012R2/lisa/stateEngine.ps1
```diff
@@ -817,49 +824,157 @@ function DoSystemDown([System.Xml.XmlElement] $vm, [XML] $xmlData)
...
+        $vm.emailSummary += "    DoApplyCheckpoint received a null xmlData parameter - VM disabled<br />"
+        $vm.currentTest = "done"
+        UpdateState $vm $Disabled
+    }
+
+    $testData = GetTestData $vm.currentTest $xmlData
+    if ($testData -is [System.Xml.XmlElement])
+    {
+        # Do not need to recover from checkpoint
+        if ($testData.noCheckpoint -and $testData.noCheckpoint -eq "True")
```
alexngmsft: Also do not recover from checkpoint when $testData.noCheckpoint is null or empty. To avoid changing behavior of existing test cases.

alexngmsft: Also do not recover from checkpoint when $testData.noCheckpoint is null or empty. To avoid changing behavior of existing test cases.

Jingli1985: Thanks for your suggestion ,alexngmsft

WS2012R2/lisa/stateEngine.ps1
```diff
@@ -817,49 +824,159 @@ function DoSystemDown([System.Xml.XmlElement] $vm, [XML] $xmlData)
...
+            {
+                $snapshotName = $vm.vmSnapshotName
+                LogMsg 5 "Info : $($vm.vmName) Over-riding default snapshotName from VM section to $snapshotName"
+            }
+
+            #
+            # Find the snapshot we need and apply the snapshot
+            #
+            $snapshotFound = $false
+            $snaps = Get-VMSnapshot $vm.vmName -ComputerName $vm.hvServer
```
alexngmsft: You should use "Get-VMSnapshot -Name $snapshotName  -VMName $vm.vmName". This will avoid the need for the foreach loop in the next part.

WS2012R2/lisa/stateEngine.ps1
```diff
@@ -817,49 +824,159 @@ function DoSystemDown([System.Xml.XmlElement] $vm, [XML] $xmlData)
...
+        UpdateState $vm $Disabled
+    }
+
+    $testData = GetTestData $vm.currentTest $xmlData
+    if ($testData -is [System.Xml.XmlElement])
+    {
+        # Do not need to recover from checkpoint
+        if (-not $testData.RevertDefaultSnapshot -or $testData.RevertDefaultSnapshot -eq "False")
+        {
+            LogMsg 9 "Info : noCheckpoint is not configured or set to True."
```
alexngmsft: Update the log message so that it mentions "RevertDefaultSnapshot" instead of "noCheckpoint"

Jingli1985:
Hi @chvalean ,
Any update for this patch, Thanks?

PR #381, opened 2016-07-06
New Manual Memory Hot Add test case

adriansuhov:
Verify that memory changes with non 128-MB aligned values. This test
will start a vm with 4000 MB and will hot add 1000 MB to it. After that,
will reboot the vm and hot add another 1000 MB. Test will pass if all
hot add operations work.


Jingli1985:
@adriansuhov 
Which platform can run those hot add memory cases?  Is Win 2016 TP5 OK?

adriansuhov:
Yes, the feature should work there

PR #392, opened 2016-07-12
Fixed ethtool script

OvidiuRusu:
Removed distro verification and added some exit codes in case of fail.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/NET_ETHTOOLS.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/NET_ETHTOOLS.sh
@@ -42,15 +42,19 @@
 #
 #############################################################################################################
 
+declare __iface_ignore
+declare __synth_iface
+declare -a ETHTOOL_KEYS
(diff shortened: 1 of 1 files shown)


WS2012R2/lisa/remote-scripts/ica/NET_ETHTOOLS.sh
```diff
@@ -129,32 +99,32 @@ if [ "${ipv4:-UNDEFINED}" = "UNDEFINED" ]; then
...
 
 	# Get the interface associated with the given ipv4
 	__iface_ignore=$(ip -o addr show | grep "$ipv4" | cut -d ' ' -f2)
 fi
 
-# Retrieve synthetic network interfaces
-GetSynthNetInterfaces
-
 declare -a ETHTOOL_KEYS
 ETHTOOL_KEYS=("tcp-segmentation-offload" "scatter-gather" "rx-checksumming" "tx-checksumming")
```
chvalean: variables declarations would go to the beggining of the script

PR #397, opened 2016-07-19
Add one new STOR_VHDX case to plug/unplug two scsi disks continuesly,…

Jingli1985:
… and minor update Add-VHDXForResize to refine ide support


chvalean:
Thanks Jing Li for the automation updates!
In regards to the new test case - VHDx_DiskHotAddRemove_Loop - we actually already have an automation in test case called VHDx_DiskHotPlugUnplug, which does the same thing.
The only difference might be the loop, which is a nice addition.

If no other high level changes are between these 2 test cases, I would propose to update the existing test case VHDx_DiskHotPlugUnplug with your additions, so that we don't have different test cases that cover the same functionality.
Thank you!

lixuemin2016:
Hi Chris,
Thank you so much for your review, will update based on comment.
To avoid duplication, plan to remove original files STOR_unPlug.sh, and STOR_unPlug.ps1, then rename new file "STOR_unPlug_Plug_Loop.ps1" to  "STOR_unPlug_Plug.ps1" which support LoopCount", and only keep one case. Is it ok?

chvalean:
I would go with STOR_unPlug_Plug.ps1 instead of STOR_unPlug_Plug_Loop.ps1 but the plan sounds good, as long as the initial script/test case functionality is kept in the loop'ed version as well.

chvalean:
@Jingli1985 I think we're missing STOR_unPlug_Plug.ps1 now after the last changes

Jingli1985:
@chvalean The file has been pushed . sorry for missing it

PR #398, opened 2016-07-19
Added CLImageStorDir variable for lisa.ps1

adriansuhov:
lisa.ps1 will now accept a new argument, CLImageStorDir, which is
represented in XML files by imageStoreDir tag. If CLImageStorDir parameter
is given to lisa:
- it will check if path exists
- it will check if XML file given has this tag in first place. If the
XML doesn't have this tag, lisa won't change anything
- if imageStoreDir tag is present in XML file, it will be overwritten
with the value given at runtime


nickme:
Would you consider changing the command line argument name to CLImageStoreDir so it more closely matches the name used in the .xml file?

adriansuhov:
@nickme Changes made

PR #400, opened 2016-07-26
Add one STOR_VHDX test case for adding and removing multiple disk, e.g. 64

Jingli1985:


WS2012R2/lisa/xml/STOR_VHDX.xml
```diff
@@ -178,6 +179,22 @@
            <timeout>800</timeout>
        </test>
 
+       <test>
+           <testName>VHDx_DiskHotPlugUnplug_Multi_SCSI</testName>
+           <preTest>
+               <file>setupscripts\AddVhdxHardDisk.ps1</file>
```
OvidiuRusu: This should go as a setupScript otherwise it will give a lot of errors on Remove-VMDvdDrive. All other scripts are using the AddVhdxHardDisk.ps1 as setupScript.

Jingli1985: Hi @OvidiuRusu , 
This script is used to test hot add and remove Disk  64 times, so we add these two script in pretest.

OvidiuRusu: Hi @Jingli1985 ,
If I run AddVhdxHardDisk.ps1 as preTest script I'm getting these kind of messages: "Remove-VMDvdDrive : The operation cannot be performed while the virtual machine is in its current state." because the VM is running. If we want to keep it like this we need a handle on Get-VMDvdDrive and Remove-VMDvdDrive command.

chvalean:
Jing, when a new test cases is created, we usually run the full xml for that test area, in order to ensure that we don't conflict with the rest of the test cases or introduce errors, like it seems to be in this case.

Jingli1985:
Chvalean , I just checked single case if there is an error and passed in general and pushed upstream to wait for merging and then run all cases in xml.  I will check this case why it has such an error like this.  
Thanks

Jingli1985:
Hi Chris, 
We fixed this issue due to no DVD drive on our VMs, meanwhile we noticed also use the AddVhdxHardDisk.ps1 in the pretest phase of some cases and fixed it together.  I guess there is an error like this when run other cases because there is just error logs in pretest phase mo any handle to stop current case .The case is still successful if the pretest script is failed

PR #404, opened 2016-08-01
convert time of VM with windows system time format

Jingli1985:
Met errors below when cases are ran on different server:
Windows time: 1/8/2016 8:16:14
Unix time:  8/1/2016 8:16:14
so need to get Unix time with windows system time format

### DIFF
--- a/WS2012R2/lisa/setupscripts/INST_LIS_TimeSync.ps1
+++ b/WS2012R2/lisa/setupscripts/INST_LIS_TimeSync.ps1
@@ -159,8 +159,8 @@ function GetUnixVMTime([String] $sshKey, [String] $ipv4)
     }
 
     #
-    # now=`date "+%m/%d/%Y %H:%M:%S%p"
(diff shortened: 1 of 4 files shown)


WS2012R2/lisa/setupscripts/INST_LIS_TimeSync.ps1
```diff
@@ -160,13 +160,13 @@ function GetUnixVMTime([String] $sshKey, [String] $ipv4)
 
     #
     # now=`date "+%m/%d/%Y %H:%M:%S%p"
-    # returns 04/27/2012 16:10:30PM
+    # returns 04/27/2012/16:10:30
```
chvalean: The previous comment was correct, on a RHEL 7.x the returned format from "+%m/%d/%Y %H:%M:%S%p" is "08/01/2016 04:13:04AM", so the PM/AM format appears.

PR #408, opened 2016-08-03
fix variable is not set

Jingli1985:


WS2012R2/lisa/xml/STOR_VHDX.xml
```diff
@@ -76,6 +76,7 @@
                 <param>DefaultSize=3TB</param>
                 <!-- <param>Path=G:\VHDs</param> -->
                 <param>fileSystems=(ext4 xfs btrfs)</param>
+                <param>ControllerType=SCSI</param>
```
chvalean: The controller type is already defined in the other param - SCSI/IDE=ID,port,sector_size,disk_size.
Why are we introducing a new parameter for ControllerType?

Jingli1985:
Actually,  The parameterof SCSI/IDE=ID,port,sector_size,disk_size already isn't used in code after updating code several times. so should delete this.

PR #411, opened 2016-08-04
Add two hyperv-daemons test cases to check default files and status

Jingli1985:
Add two cases of LIS daemons
1. Check the related file of daemons if it exist on RHEL7 VM
2. Check the related service of daemons if it is active when system start on RHEL7 VM


chvalean:
Thanks for the PR JingLi, these checks on the LIS daemons are welcomed.

I do have a few suggestions on them:
1. Merge them into a single test case, verify the presence of the daemon binaries, then the enable status.
2. Have this as a Core test, not a separate xml.
3. The 'Guest Services' integration service must be verified to be enabled or have it enable as part of a setup script. This is similar to what we're already doing for the FCopy test cases.
4. Modularize the test cases in a per distro basis, as we would then have to further work on this to make it support other distros as well.

Thank you!

chvalean:
It seems that a failure for any of the 3 daemons will hang the test and not exit/report the failure.
This also occurs if there is an external LIS version (4.x) installed on RHEL 7.x for example. The daemons are running, but the automation fails to find the service file.

In the VM script log there would be this message, however the PowerShell side of the script will not recognize the failure and the test case will continue to run, although the test failed.
Would you please be able to remedy the failure situation?

`Tue Aug 09 13:35:38 2016 : ERROR: hypervfcopyd.service is not in /usr/lib/systemd or /etc/systemd/system/multi-user.target.wants , test failed`

lixuemin2016:
Hi Chris,

Thank you so much for your review comment, I have reproduced this hang issue when case failure. 
When calling RunRemoteScript for Hyperv_Daemons_Files_Status.sh, it is failed to pull state.txt from VM to host, will update change later.
Thank you.

Jingli1985:
Hi Chris,
We do mirror fix again and can't reproduce the error log above `Tue Aug 09 13:35:38 2016 : ERROR: hypervfcopyd.service is not in /usr/lib/systemd or /etc/systemd/system/multi-user.target.wants , test failed`, 
Do you do some change on your VM? Anyway, Could you please to make sure the hypervfcopyd is enabled by default? While you enable the service, a symbolic link is placed in the /etc/systemd/system/multi-user.target.wants directory or else it is not, The daemons are enabled by default on RHEL.

chvalean:
I cannot merge this because of some conflicts, most likely in the Core XML as we have another test ID 30 introduced.

chvalean:
Still getting a conflict here, I tried to find out the culprit lines but couldn't identify them.

Jingli1985:
Hi Chris, 
Do an new PR instead. Thanks

PR #413, opened 2016-08-04
Add Hyperv-daemons selinux check by ausearch

Jingli1985:


chvalean:
Thank you JingLi, we've taken these and implemented in request https://github.com/LIS/lis-test/pull/422

PR #417, opened 2016-08-08
fix undefine parameter and update xml file

Jingli1985:


chvalean:
Thank you for the changes!
Running the STOR VHDx XML (excluding NFS test cases), from a total of 23 tests:
- 9 passed successfully
- 3 tests aborted
- 11 failed - all due to setup script failed

This has been verified with a RHEL 7.2 VM.
From what I see, most of the failures are with this message:

> Error: Invalid SCSI/IDE arguments, only support to define one SCSI disk

Jingli1985:
It is caused by defining $diskcount=1 but not $null in RemoveVHDXHardDisk.ps1, which makes fail to remove disk.

WS2012R2/lisa/xml/STOR_VHDX.xml
```diff
@@ -114,6 +116,7 @@
                 <param>DefaultSize=64TB</param>
                 <!-- <param>Path=G:\VHDs</param> -->
                 <param>ControllerType=SCSI</param>
+                <param>SCSI=0,0,Dynamic,512,64TB</param>
```
chvalean: why do we require this line if all the other params have been moved outside of it and individually?

Jingli1985: this param is used to Count the number of SCSI= and IDE= entries in constants in /STOR_Large_Disk.sh

chvalean:
Re-running all STOR VHDx test cases, I had a successful complete run!

PR #425, opened 2016-08-17
fix error ID when removing DVD with gen2 VM

Jingli1985:


WS2012R2/lisa/setupscripts/RemoveIsoFromDvd.ps1
```diff
@@ -87,13 +88,30 @@ $error.Clear()
 #
 # Remove the .iso file from the VMs DVD drive
 #
+
+#
+#Get the version of VM
+#
+$vmGeneration = Get-VM $vmName | select -ExpandProperty Generation
```
chvalean: Please use the -ComputerName param for Get-VM.
This is required if the VM is on a remote server.

WS2012R2/lisa/setupscripts/RemoveIsoFromDvd.ps1
```diff
@@ -87,13 +88,30 @@ $error.Clear()
...
+{
+    $controllerNumber=0
+}
+
 $dvdcount = $(Get-VMDvdDrive -VMName $vmName).ControllerLocation.count 
-for ($i=0; $i -lt $dvdcount; $i++)
+for ($i=0; $i -le $dvdcount; $i++)
 {
-    $dvd = Get-VMDvdDrive -VMName $vmName -ControllerNumber 1 -ControllerLocation $i
+    $dvd = Get-VMDvdDrive -VMName $vmName -ControllerNumber $controllerNumber -ControllerLocation $i
```
chvalean: Same here for using -ComputerName, I see that in the older code it wasn't either specified.

PR #431, opened 2016-08-24
Adding performance prep script to customize sysctl and i/o disk scheduler

mihaico:
…


chvalean:
Bash scripts reside under remote-scripts/ica folder, so please move this under that path.

Also, the Apache license must be added.

PR #478, opened 2016-09-29
DM: switched to stress-ng as memory pressure tool

zsoltdudas:
- Switched to stress-ng as memory pressure tool from stresapptest
- The only remaining test using stressapptest is HotAdd. It can be used for a while to double check the validity of stress-ng
- Changed the way stress tests handle memory and pressure calculations
- Tested on SLES 12 and partially on RHEL 7.2. Should work well with kernels 3.x and 4.x

Please report any issues or misbehavior as these are not only guest, but host dependent as well.

### DIFF
--- a/WS2012R2/lisa/setupscripts/DM_HighPriority.ps1
+++ b/WS2012R2/lisa/setupscripts/DM_HighPriority.ps1
@@ -74,7 +74,7 @@ function checkStressapptest([String]$conIpv4, [String]$sshKey)
 
     $cmdToVM = @"
 #!/bin/bash
-        command -v stressapptest
+        command -v stress-ng
         sts=`$?
         if [ 0 -ne `$sts ]; then
             echo "Stressapptest is not installed! Please install it before running the memory stress tests." >> /root/HotAdd.log 2>&1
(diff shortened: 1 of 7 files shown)


WS2012R2/lisa/setupscripts/DM_HotAdd.ps1
```diff
@@ -72,7 +72,7 @@ function checkStressapptest([String]$conIpv4, [String]$sshKey)
 
     $cmdToVM = @"
 #!/bin/bash
-        command -v stressapptest
+        command -v stress-ng
```
adriansuhov: No need for this, stressapptest is still used here

WS2012R2/lisa/setupscripts/DM_HighPriority.ps1
```diff
@@ -74,7 +74,7 @@ function checkStressapptest([String]$conIpv4, [String]$sshKey)
 
     $cmdToVM = @"
 #!/bin/bash
-        command -v stressapptest
+        command -v stress-ng
         sts=`$?
         if [ 0 -ne `$sts ]; then
             echo "Stressapptest is not installed! Please install it before running the memory stress tests." >> /root/HotAdd.log 2>&1
```
adriansuhov: Stress-ng insdead of Stressapptest

chvalean:
Closing PR until further changes

PR #493, opened 2016-10-18
Add test cases to check MAX NIC support

bogdancarpusor:


bogdancarpusor:
[add TCs to verify max NICs support](https://trello.com/c/yUjr5Tpd/399-add-tcs-to-verify-max-nics-support)

PR #502, opened 2016-11-10
Add support for specific port addresses on FC tests

bogdancarpusor:
Now you can specify different port addresses for fibre channel tests
through parameters in an xml file - WWNN=portA,portB - WWPN=portA,portB

If the specific addresses are already in use by some other local VM
the test will fail.

### DIFF
--- a/WS2012R2/lisa/setupscripts/FC_AddFibreChannelHba.ps1
+++ b/WS2012R2/lisa/setupscripts/FC_AddFibreChannelHba.ps1
@@ -40,7 +40,7 @@
             <param>vSANName=fc</param>
         </testParams>
     </test>
-    
+
 .Parameter vmName
     Name of the VM to perform the test with.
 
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/setupscripts/FC_AddFibreChannelHba.ps1
```diff
@@ -122,11 +131,35 @@ if (-not $vSANName) {
 
 # Add the FC adapter, if the command is successful there is no output
 Write-Output "Adding the Fibre Channel adapter..."
-if ((Add-VMFibreChannelHba -VmName $vmName $vSANName) -ne $null) {
+if ((Add-VMFibreChannelHba -VmName $vmName -SanName $vSANName)-ne $null) {
```
chvalean: please add the ComputerName param as well

WS2012R2/lisa/xml/FC_Tests.xml
```diff
@@ -123,6 +118,9 @@
             <ipv4></ipv4>
             <sshKey>PKI_id_rsa.ppk</sshKey>
             <suite>FC</suite>
+            <testParams>
+                <param>vSANName=FC_NAME</param>
```
chvalean: please add a commented version of the WWNN/WWPN params for reference

PR #523, opened 2016-12-15
Added debug file for test cases that fail

bogdancarpusor:
A new debug file was added that covers more details
regarding the state of a VM for a failing test case

The feature is enabled during LISA runs by setting the
<collect> parameter as True


chvalean:
[analyze the option of jenkins artifacts for failed jobs](https://trello.com/c/9oiMmjgB/428-analyze-the-option-of-jenkins-artifacts-for-failed-jobs)

PR #524, opened 2016-12-15
Correct syntax error when referencing evevent ID

nickme:


WS2012R2/lisa/setupscripts/STOR_VSS_BackupRestore_Fail.ps1
```diff
@@ -304,7 +304,7 @@ if(-not $EventLog)
 # Event ID 10107 is what we looking here, it will be always be 10107.
 foreach ($event in $EventLog)
    {
-       Write-Output "VSS Backup Error in Event Log number is $EventLog:" >> $summaryLog
+       Write-Output "VSS Backup Error in Event Log number is $($event.ID):" >> $summaryLog
        if ($event.Id -eq 10150)
```
Jingli1985: Hi Nick,
Which event ID will be captured on host? here are different ID checked in comments and codes(10107 and 10150), In fact, i can't capture the both ID when i test this case, please help me to check this  case

Thanks

chvalean: Hi Jing Li,

For this test case we're trying to cause the backup operation to fail. When that happens, event ID 10107 is recorded.
Now this is a hit or miss, depending on multiple factors. We don't have a fully reliable method to trigger that failure at the exact time for the backup to fail, so for that reason you might not see it all the time.

PR #525, opened 2016-12-16
add some checking point

Jingli1985:
Here are some scenario need to be checked
1. make sure which cd drive mounted an ISO
2. Adjust time to wait for VM started completely
3. Sometimes, The VM will disappear if the restore failed when do the cases related to backup&restore cases, so need to check if VM exist in shutting VM

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
@@ -97,12 +97,13 @@ if [ $os_VENDOR != "Ubuntu" ] && [ $os_VENDOR != "Debian" ]; then
 fi
 sleep 1
 LogMsg "Mount the CDROM"
-for drive in $(ls /dev/cdrom*)
+for drive in $(ls /dev/sr*)
 do
     blkid $drive
     if [ $? -eq 0 ]; then
         mount -o loop $drive /mnt/
         LogMsg "Mount the CDROM ${drive}"
+        break
     fi
 done    
 sts=$?
--- a/WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
@@ -97,7 +97,14 @@ if [ $os_VENDOR != "Ubuntu" ] && [ $os_VENDOR != "Debian" ]; then
 fi
 sleep 1
 LogMsg "Mount the CDROM"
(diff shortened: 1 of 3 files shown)


WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
```diff
@@ -97,7 +97,14 @@ if [ $os_VENDOR != "Ubuntu" ] && [ $os_VENDOR != "Debian" ]; then
 fi
 sleep 1
 LogMsg "Mount the CDROM"
-mount /dev/cdrom /mnt/
+for drive in $(ls /dev/cdrom*)
+do
+    blkid $drive
```
paulaCrismaru: You should make sure /mnt/ is not mounted.
In case you have more than 1 CDs attached to the VM, the CD to be mounted will not be the one used for the test. The test will use the first CD found, since the other attempts to mount in /mnt/ will fail.

WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
```diff
@@ -97,7 +97,14 @@ if [ $os_VENDOR != "Ubuntu" ] && [ $os_VENDOR != "Debian" ]; then
 fi
 sleep 1
 LogMsg "Mount the CDROM"
-mount /dev/cdrom /mnt/
+for drive in $(ls /dev/cdrom*)
```
paulaCrismaru: If you have more than 1 CD attached, the devices will be found at /dev/sr*

Jingli1985:
Wait for VM started completely , the 200s is not enough for RHEL6 

发自我的 iPhone

Jingli1985:
Yes only an ISO is mounted in my VM but I am sure which CD drive it is mounted ， sometimes there are more than one CD drive attached in my VM

发自我的 iPhone

WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
paulaCrismaru: The test scenario is to see if data can be read from a CD. My opinion is that you can stop after the first CD (/dev/cdrom) is mounted (as in the deleted line 100).

WS2012R2/lisa/remote-scripts/ica/Core_LIS_CD.sh
Jingli1985: Thanks, I will update it.

PR #537, opened 2017-01-12
SR-IOV - Improvements to old TCs

adriansuhov:
Old TCs are now using functions from SR-IOV_Utils.sh. Duplicate code was
significantly reduced.
A cleanup script was also added for turning off the dependency VM or VMs
where is the case.


adriansuhov:
[SR-IOV new test cases & improvements](https://trello.com/c/bxdstO2V/437-sr-iov-new-test-cases-improvements)

PR #539, opened 2017-01-12
Add new test case for net corruption bug

bogdancarpusor:
The following automation transfers a file from a linux VM
to a WS host using netcat. It checks for call traces,
during the transfer process, and also compares the
file hashes in order to validate the transfer result.

### DIFF
--- a/WS2012R2/lisa/setupscripts/NET_Corruption.ps1
+++ b/WS2012R2/lisa/setupscripts/NET_Corruption.ps1
@@ -1,0 +1,256 @@
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the ""License"");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


WS2012R2/lisa/setupscripts/NET_Corruption.ps1
```diff
@@ -0,0 +1,248 @@
...
+if (-not $?)
+{
+    $msg = "Error: Unable to get file hash from VM"
+    Write-Output $msg | Tee-Object -Append -file $summaryLog
+    return $False
+}
+
+"Info : File hashes: ${remoteHash} - ${localHash}"
+if ($remoteHash.ToUpper() -ne $localHash) {
+    $msg = "Error: File hashes do not match."
```
chvalean: I'm always receiving this message, even on latest upstream kernel where this issue is fixed.
Also verified with 1% corruption and exit code remains, it seems to be from automation.

bogdancarpusor: Netcat binary was missing from \bin folder and no error was raised for this case. I've added a check, at the beginning of the test script, to verify if the path ".\bin\nc.exe" actually exists.

PR #589, opened 2017-02-28
Migrate xiaofeng's ESX-lisa commit to create xml report by updating O…

lixuemin2016:
…SAbstractions, stateEngine, utilFunctions

Change-Id: Ibab9121d33bf4e12dc0db1b29fa762f5cd39b280


lixuemin2016:
1. Add JUnit XML result file support.
    there'll be a new Report-.xml file in the case result folder.
    result xml file follows JUnit rules and is acceptable by Jenkins and other CI applications.
2. Add case running time support.
    this feature finds how long the test case will last.
    the running time will be included in result XML file for each test case.
    the running time will be calculated from system down to system down if case needs reboot, or from system up to determinreboot if case doesn't need reboot.

PR #615, opened 2017-03-15
Fix module_check for rhel6 without hv_netvsc by default and update xm…

lixuemin2016:
…l to include more modules

1. For rhel6, by default without hv_netvsc module except install vm by network, so need to install dracut-network, then "yum install -y dracut-network,  dracut -f" to make it available, refer to https://bugzilla.redhat.com/show_bug.cgi?id=1298445 
2. Update hid-hyperv.ko and hyperv_fb.ko to the modules list.

### DIFF
--- a/WS2012R2/lisa/xml/CoreTests.xml
+++ b/WS2012R2/lisa/xml/CoreTests.xml
@@ -292,7 +292,7 @@
 			<files>remote-scripts/ica/modules_check.sh</files>
 			<testparams>
 				<param>TC_COVERED=CORE-24</param>
-				<param>hv_modules=(hv_vmbus.ko hv_storvsc.ko hv_netvsc.ko hyperv-keyboard.ko)</param>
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/xml/CoreTests.xml
```diff
@@ -292,7 +292,7 @@
 			<files>remote-scripts/ica/modules_check.sh</files>
 			<testparams>
 				<param>TC_COVERED=CORE-24</param>
-				<param>hv_modules=(hv_vmbus.ko hv_storvsc.ko hv_netvsc.ko hyperv-keyboard.ko)</param>
+				<param>hv_modules=(hid-hyperv.ko hv_vmbus.ko hv_storvsc.ko hv_netvsc.ko hyperv_fb.ko hyperv-keyboard.ko)</param>
```
chvalean: Thank you for the addition on dracut-network!

For this test case we are looking at the minimal components that would ensure a Linux VM continues to work.
hid-hyper-v - mouse input and hyperv_fb are not mandatory components, so for that reason we have not included them.

lixuemin2016:
Updated based on comments, thank you.

PR #618, opened 2017-03-17
Minor update KvpTests to add pretest for kvp_remove_key_value

lixuemin2016:
Add pretest for kvp_remove_key_value,  <PreTest>setupScripts\AddKeyValue.ps1</PreTest>, in order to reduce the dependency between test cases, even set noReboot as True.

When test locally we set noReboot as false and revert snapshot for every case, need to add pretest script to add key value, then remove key value.

### DIFF
--- a/WS2012R2/lisa/xml/KvpTests.xml
+++ b/WS2012R2/lisa/xml/KvpTests.xml
@@ -64,7 +64,7 @@
                 <param>TC_COVERED=SQM-01</param>
             </testparams>
         </test>
-        
+
         <test>
             <testName>KVP_Basic</testName>
             <testScript>SetupScripts\KVP_Basic.ps1</testScript>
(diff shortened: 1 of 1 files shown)


lixuemin2016:
Maintain it in local repo, do not change in upstream.

chvalean:
@lixuemin2016 the kvp tests are designed to run without any reboot operations, thus saving time on the testing duration.
Do you require reboot operations between these tests?

lixuemin2016:
Hi Chris, 
Understand kvp tests are designed to save time without any reboot.
Locally in order to make test case independent and easy to debug, we set noReboot as false and revert snapshot for every case, so need to add pretest script to add key value, then remove key value. It is ok if it is different with upstream for xml config, have maintained locally.
Thank you for your comments.

Issue #572, opened 2017-02-16
Fix VHDx storage tests compatibility with nano / remote server


chvalean:
Running the STOR VHDx tests on a remote Nano Server results in several failures due to missing remote capabilities of the scripts.
This should start from RemoveVhdxHardDisk.ps1, other related scripts might require improvements as well.

PR #636, opened 2017-03-27
Fixed remote capability for Stor_VHDX tests on Nano Server

ilenghel:
Added minor changes to fix remote capabilities for Stor_VHDX tests on Nano Server


chvalean:
Addressing issue https://github.com/LIS/lis-test/issues/572

PR #638, opened 2017-03-29
Add VHDX IDE skipped if vm is gen2, cover AddVhdxHardDisk, Add-VHDXFo…

lixuemin2016:
…rResize, related remove and resize scripts

Mainly used to skip test cases for gen 2 VM adding IDE cases. 

For setup script, return $True if need to skip when add IDE to gen2 VM, but related test script needs to return "Skipped"
For test script, return $Skipped if need to skip directly.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/STOR_Lis_Disk.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/STOR_Lis_Disk.sh
@@ -179,9 +179,6 @@ function TestFileSystem()
             fi
          fi
     done
-
-
-
 }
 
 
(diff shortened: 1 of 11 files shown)


chvalean:
Thank you for these improvements @lixuemin2016 !

Would be much appreciated if you can make the below small changes:
1. there is a typo in the script name "#Source **TCUitls**.ps1"
2. for the text that reads "vm generation 2" - the correct definition would be "Generation 2 VM" as it's commonly agreed for this type.

lixuemin2016:
Chris, thank you so much for your review comments. I have updated scripts based on your comments.

PR #649, opened 2017-04-05
Update NUMA to cover test for kernel 2.6.*

lixuemin2016:


chvalean:
On RHEL 6.4 (kernel 2.6.32-358) we don't see the NUMA nodes spreading, so there it is not available.
Can you please check and confirm from which kernel NUMA support was introduced and maybe add that version up to where we would do the skip in the automation?
Thank you!

lixuemin2016:
@chvalean 
Have updated based on your review comments.
NUMA feature is enabled in RHEL6.5, based on https://access.redhat.com/solutions/802693, 
but there is performance regression on NUMA systems in RHEL-6.5, so set  NUMA supported version from RHEL6.6.

chvalean:
We're looking at consolidating code and not having duplicate functions in multiple scripts.
As the 2 PS scripts modified already source TCUtils, can you please move function GetNumaSupportStatus in there?
Thank you also for clarifying the NUMA support!

lixuemin2016:
@chvalean 

Have updated based on your comment.  Agree with you, yesterday I also hesitated to move it to NUMAUtil.PS1 or TCUtils.PS1.

Thank you so much.

PR #660, opened 2017-04-27
Combine NMI_Send_Interrupt to NMI_Verify_Interrupt and update NMI_Verify_Interrupt

HuijingHei:
Refer to http://driverdev.linuxdriverproject.org/pipermail/driverdev-devel/2016-November/097202.html:
Prior to WS2016 the NMI is injected to all CPUs of the guest and WS2016 injects it to CPU0 only.

Update:
1. Remove NMI_Send_Interrupt, combine it to NMI_Verify_Interrupt
2. When host version is older than WS2012R2, return skipped
3. When host version is WS2012R2, check the NMI is injected to all CPUs of the guest
4. When host version is WS2016, check the NMI is only injected to CPU0

### DIFF
--- a/WS2012R2/lisa/setupscripts/TCUtils.ps1
+++ b/WS2012R2/lisa/setupscripts/TCUtils.ps1
@@ -1550,7 +1550,7 @@ function GetVMGeneration([String] $vmName, [String] $hvServer)
     #>
     $vmInfo = Get-VM -Name $vmName -ComputerName $hvServer
 
-    # Hyper-v server 2012 only supports generation 1 VM.
+    # Hyper-V Server 2012 (no R2) only supports generation 1 VM
     if ( $vmInfo.Generation -eq "")
     {
         $vmGeneration = 1
@@ -1566,15 +1566,14 @@ function GetVMGeneration([String] $vmName, [String] $hvServer)
 #
 # GetNumaSupportStatus()
 #
-##########################################################
-#############
+#######################################################################
 function GetNumaSupportStatus([string] $kernel)
 {
     <#
     .Synopsis
-        Try to determine whether guest supports numa
+        Try to determine whether guest supports NUMA
     .Description
-        Get whether numa is supported or not based on kernel verison.
+        Get whether NUMA is supported or not based on kernel verison.
         Generally, from RHEL 6.6 with kernel version 2.6.32-504,
         NUMA is supported well.
     .Parameter kernel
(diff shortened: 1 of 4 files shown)


WS2012R2/lisa/setupscripts/TCUtils.ps1
```diff
@@ -1582,23 +1583,59 @@ function GetNumaSupportStatus([string] $kernel)
...
-        return $false
+    if( $kernel.Contains("i686") `
+        -or $kernel.Contains("i386")){
+            return $false
     }
+    $numaSupport = "2.6.32.504"
+    $kernelSupport = $numaSupport.split(".")
+    $kernelCurrent = $kernel.replace("-",".").split(".")
 
-    if ( $kernel.StartsWith("2.6")) {
```
chvalean: We need to keep this code as we're checking only if we have the 2.6 kernel for NUMA support, otherwise we don't want to process the kernel build

HuijingHei: OK, thanks. Maybe I did not pull before commit. Have updated.

chvalean:
Thank you @HuijingHei, it was on my list to update NMI for a while now.

While reviewing the code I was considering that we should not exclude WS2012R2.
I'm thinking more of a condition like:
- if WS2016 check that the NMI spread goes only to CPU0
- if WS2012R2 then the passing criteria is that all CPUs will get the NMI.
- if OS build number is lower than 9600 only then Skip the test.

Please let me know your feedback on this workflow and if you can accommodate the changes.
Thank you!

HuijingHei:
Hi @chvalean ,
Thanks for your feedback!
Yes, these commits have included what you considered.

    if OS build number is lower than 9600 only then Skip the test.
-   NMI_Verify_Interrupt.ps1 

      if BuildNumber lower than 9600; return $Skipped
      else send $BuildNumber to vm's constants.sh

    if WS2016 check that the NMI spread goes only to CPU0
    if WS2012R2 then the passing criteria is that all CPUs will get the NMI.
-   nmi_verify_interrupt.sh

       if cpu0 || 2012R2; then check cpu get NMI(BuildNumber is lower than 14393, means host is 2012R2)
       else check cpu not get NMI (not cpu0 and not 2012R2)

Additional info: 
	I combine NMI_Send_Interrupt to NMI_Verify_Interrupt, because I want to send $BuildNumber to vm's constants.sh

Thanks
Huijing Hei

HuijingHei:
Hi @chvalean ,
Thanks for your feedback!

Yes, these commits have included what you considered.

Thanks
Huijing Hei

PR #666, opened 2017-05-02
Add Stor VHDXResize skipped if host version lower than 9600

lixuemin2016:
Add skipped if host version is lower than 2012R2, even already have <dependency> in the case STOR_VHDXResize.xml file, currently it cannot return skip, locally(downstream) we run same xml file for all the cases, so open this pull request.

Thank you

### DIFF
--- a/WS2012R2/lisa/setupscripts/STOR_VHDXResize_Basic.ps1
+++ b/WS2012R2/lisa/setupscripts/STOR_VHDXResize_Basic.ps1
@@ -151,6 +151,18 @@ else
     return $false
 }
 
+# if host build number lower than 9600, skip test
+$BuildNumber = GetHostBuildNumber $hvServer
+
+if ($BuildNumber -eq 0)
+{
+    return $false
+}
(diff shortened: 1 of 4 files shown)


chvalean:
Hi Li,
We already have a platform check in the [VHDx Resize xml here](https://github.com/LIS/lis-test/blob/master/WS2012R2/lisa/xml/STOR_VHDXResize.xml#L28)

lixuemin2016:
@chvalean 
Yes, have notice that already have supported platform check,  <hostVersion>6.3.9600</hostVersion>, locally we run all the cases in one xml for all the hosts/guests, so could not add this section to global section as VHDx Resize.xml.

If add <hostVersion> under <test>, need to add more code to lisa framework, if you consider that no need to update resize script in the upstream, also agree that, I will abandon this pull request.  Only one thing a little inconvenient we need to update test results from "Failed" to "Skipped" for 2012 which have automatically entered to Polarion(test case management tool) .
Thank you so much.

PR #678, opened 2017-05-12
Skip hypervvssd and hypervfcopyd check for hyperv 2012 host

lixuemin2016:
Hyper-v 2012 host does not support backup and fcopy feature, in the vm side, even related file available in the related file, here also ignore check, only verify hypervkvpd status.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/Hyperv_Daemons_Files_Status.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/Hyperv_Daemons_Files_Status.sh
@@ -84,37 +84,46 @@ LinuxRelease()
 }
 
 #######################################################################
-# Check hyper-daemons default files and service status
+# Check hyper-daemons default files and service status.
+# If BuildNumber < 9600, hypervvssd and hypervfcopyd service are inactive,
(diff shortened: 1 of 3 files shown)


WS2012R2/lisa/remote-scripts/ica/Hyperv_Daemons_Files_Status.sh
```diff
@@ -84,37 +84,46 @@ LinuxRelease()
...
+    "RHEL7" | "CENTOS7")
+        for (( i=0; i<$len_hv; i++))
+        do
+          CheckDaemonsFilesRHEL7 ${hv_service[$i]}
+          CheckDaemonsStatusRHEL7 ${hv_service[$i]}
+        done
+        ;;
+    *)
+        LogMsg "Distro not supported"
+        UpdateTestState "TestAborted"
```
chvalean: Can the test exit with TestSkipped if the distro is not HREL/CentOS?

lixuemin2016: I will check and try to skip for other distro, thank you.

PR #698, opened 2017-06-07
DM_CONFIGURE_MEMORY.ps1: Fix set vm memory failed issue; NUMA_off_ker…

HuijingHei:
1. Update DM_CONFIGURE_MEMORY.ps1:

- Set vm memory failed, because it can not get vmName from testParams, vmName para already receive the value, so delete tPvmName para

- Add verification after set vm memory

2. NUMA_off_kernel.sh: Add fedora support

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/NUMA_off_kernel.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/NUMA_off_kernel.sh
@@ -232,7 +232,7 @@ case $DISTRO in
         ConfigCentos
     ;;
 
-    "redhat_7" | "centos_7")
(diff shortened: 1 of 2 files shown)


chvalean:
Thank you for the code enhancements!

One quick question to better understand the new check for VM memory - in which situation have you come accross where $vm_mem wasn't equal with $tPstartupMem after Set-VMMemory?
If that would fail at Set-VMMemory, then the error is caught there.

HuijingHei:
Thanks for your review!

I have not come across that error.
I just met the script DM_CONFIGURE_MEMORY.ps1 failed to set vm memory, but I got no output about set vm memory failed logs, so I added this verification to locate the error.

PR #712, opened 2017-06-27
Update STOR_VSS_BackupRestore for hyper-v 2012 host and skip for IDE …

lixuemin2016:
…on gen2

Update STOR_VSS_BackupRestore for skip IDE when testing on Gen2 and skip for 2012 host. Even has xml config about dependency hostVersion config, locally we run all the cases in one xml, need to update script to make it independently.

Hi Chris,
Before we cannot run successfully on Hyper-V 2012R2 host, since after vm restore, it cannot be stopped normally,  refer to bug https://bugzilla.redhat.com/show_bug.cgi?id=1304248. It is so great latest backup script can run successfully on 2012R2 host, could you please share how it works? I did not find any clue from test script yet.

### DIFF
--- a/WS2012R2/lisa/setupscripts/STOR_VSS_3Chain_VHD_backup.ps1
+++ b/WS2012R2/lisa/setupscripts/STOR_VSS_3Chain_VHD_backup.ps1
@@ -24,20 +24,20 @@
     This script tests VSS backup functionality.
 
 .Description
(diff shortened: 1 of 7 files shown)


chvalean:
Thank you for the changes @lixuemin2016 

I will e-mail you in regards to the mentioned bug report, I don't recall seeing it.

PR #714, opened 2017-06-28
VLAN Trunking: Added skipped state for RHEL/CentOS 6.6 and lower

adriansuhov:
We will now skip VLAN Trunking TC on RHEL/CentOS 6.6 and lower because
the feature is unsupported


lixuemin2016:
Hi Adriansuhov,
Based on "Closed and verified Bug https://bugzilla.redhat.com/show_bug.cgi?id=1001424, [Hyper-V][RHEL 6.5]The RHEL guest 802.1q VLAN tagging does not work with Hyper-V in trunk mode" , VLAN Trunking starts to support from RHEL6.7.  Would you like to do minor change again?

Thank you.
Best Regards,
Xuemin

PR #724, opened 2017-07-06
Add one STOR_VSS_BackupRestore case to cover mount readonly squashfs

lixuemin2016:
Refer to bug https://bugzilla.redhat.com/show_bug.cgi?id=1160584 to add new test case.


chvalean:
I took a look at the linked bug on a RHEL 7.0 VM running the default GA kernel - and LIS built-in, though there are no messages from VSS as specified in the bug report.
The backup operation does work even with the mounted read-only volume.

Are you sure that we're exercising the root issue properly here?

WS2012R2/lisa/remote-scripts/ica/STOR_VSS_Disk_Mount_Squashfs.sh
```diff
@@ -0,0 +1,92 @@
...
+}
+
+# Source constants file and initialize most common variables
+UtilsInit
+
+# Get distro
+GetDistro
+
+case $DISTRO in
+    redhat* | fedora*)
```
chvalean: please also add centos* here

lixuemin2016:
Hi Chris,

I have just tried on one RHEL7.0 VM(not formal released version).  Yes, there is no message "Hyper-V VSS: VSS: freeze of /mnt/squash: Operation not supported", but it fails to backup vm with squashfs mounted.  Note: In normal condition, it could backup this vm successfully.

Reproduce step: 
1. Execute following commands:
yum install squashfs-tools
mksquashfs /dir dir.sqsh
mount dir.sqsh /mnt -t squashfs -o loop.
2. Backup from host, observe that vm fails to backup. 

This pull request does not cover "udf" type. From the bug comments and fix patch, looks that developer did not fix utf issue. 

Also backup xml gets conflict issue, could you please help to resolve it?

Thank you so much.

Best Regards,
Xuemin

PR #736, opened 2017-07-20
Update INST_timesync_change_state.ps1

HuijingHei:
1) Make sleep time as a parameter
Because 600s is a fixed time and can not change, so make it as a parameter can define in xml

2) Add off chrony service
In our case, we want to test the guest time can be sync to host when chrony off.

3) Add 2 test cases in xml according to the guest time can be sync to host when chrony off

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/Core_Config_TimeSync.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/Core_Config_TimeSync.sh
@@ -109,7 +109,7 @@ ConfigPtp4l()
 InstallChrony()
 {
     case $1 in
-    "CENTOS" | "RHEL")
+    "CENTOS" | "RHEL" | "FEDORA")
         yum install chrony -y
     ;;
     "UBUNTU")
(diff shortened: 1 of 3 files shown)


HuijingHei:
@chvalean I think this feature are also supported on other distros. Refer to 
https://bugzilla.redhat.com/show_bug.cgi?id=1388419
Bug 1388419 - [RFE][Hyper-V][RHEL-7.4] Add TimeSync version 4.0 protocol support

chvalean:
chrony as a service is not installed by default on any distro.
It can be installed on SLES, Ubuntu and RedHat, so we need to handle at least the installation and config phase for it before getting to this part of the automation where we disable/enable chrony.

HuijingHei:
@chvalean Thanks for your suggestion!
I have checked in INST_timesync_change_state.ps1
1) it will install and config chrony on SLES, Ubuntu and RedHat, and aborted on other distros
2) Check the host and guest time diff
3) Pasue/Save the guest, sleep xx seconds
4) Start the guest and check the diff time between host and guest

I add chrony off after step2, so I think I do not need to install chrony before disabling chrony.

PR #740, opened 2017-07-21
Update stateEngine to support xunit xml result with TC_Covered

lixuemin2016:
Migrate change from xiaofeng, add property with "TC_COVERED"

### DIFF
--- a/WS2012R2/lisa/utilFunctions.ps1
+++ b/WS2012R2/lisa/utilFunctions.ps1
@@ -62,6 +62,9 @@ function GetJUnitXML()
 <testcase name="" time="">
     <skipped/>
     <failure type=""></failure>
+    <properties>
+        <property name="TC_COVERED" value="" />
+    </properties>
 </testcase>
 </testsuite>
 '@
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/utilFunctions.ps1
```diff
@@ -253,6 +260,16 @@ function SetTestResult([String] $testName, [String] $completionCode)
             $newTestCase.failure.InnerText = "Test $testName Aborted."
         }
     }
+
+    $testID = GetTestID $testName $xmlData
+    foreach ($property in $newTestCase.properties.property)
+    {
+        if ($property.name -eq "TC_COVERED")
+        {
+            $property.value = $testID
```
chvalean: If the TC_COVERED field is missing from xml for a test case, this error is displayed by LISA at runtime:

Info :    GetTestID(Verify_LIS_Modules)
Cannot set "value" because only strings can be used as values to set XmlNode properties.
At F:\lis-test\WS2012R2\lisa\utilFunctions.ps1:269 char:13
+             $property.value = $testID
+             ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], SetValueException
    + FullyQualifiedErrorId : XmlNodeSetShouldBeAString

lixuemin2016: @chvalean Updated, if does not define TC_COVERED, output empty string to value, also change  TC_COVERED to lower case which looks better.
Thank you.

PR #748, opened 2017-08-01
update network ping ip 8.8.4.4 to 8.8.8.8

lixuemin2016:
Sometimes fails to ping 8.8.4.4 in Beijing,China, but could always ping 8.8.8.8, submit pull request to update  to 8.8.8.8, if these two address both work well in your side, how about change to 8.8.8.8?

Thank you so much.


chvalean:
In our network we had the exact opposite - we cannot ping 8.8.8.8 but instead can reach 8.8.4.4.
Given your case, we might have to look for another default IP to use and update all automation to it.

lixuemin2016:
@chvalean got it, I can update two hard codes here, e.g. $externalIP = "8.8.4.4" to xml parameter, we have our local xml files, then change it locally, thank you.

PR #750, opened 2017-08-03
Added new TC to StressTests suite

ilenghel:
Added a test case to boot the VM with more than 67GB of RAM and check
how much is accessible by VM.

### DIFF
--- a/WS2012R2/lisa/setupscripts/STRESS_BootLargeMemory.ps1
+++ b/WS2012R2/lisa/setupscripts/STRESS_BootLargeMemory.ps1
@@ -1,0 +1,124 @@
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved.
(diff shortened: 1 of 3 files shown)


WS2012R2/lisa/setupscripts/STRESS_BootLargeMemory.ps1
```diff
@@ -0,0 +1,111 @@
...
+# Create log file
+$summaryLog = "${vmName}_summary.log"
+del $summaryLog -ErrorAction SilentlyContinue
+Write-Output "This script covers test case: ${TC_COVERED}" | Tee-Object -Append -file $summaryLog
+
+#Define peak available memory in case of problem
+$peakFaultMem = [int]67700
+
+#Get VM available memory
+$guestReadableMem = bin\plink.exe -i .\ssh\$sshKey root@$ipv4 "free -m | grep Mem | xargs | cut -d ' ' -f 2"
```
chvalean: we should also add a check in /proc/meminfo to compute the memory

PR #752, opened 2017-08-07
Update InjectIPconstants default server as 8.8.4.4 and support parameter

lixuemin2016:
add one parameter about remote_sever to support change remote server ip, 8.8.4.4 could not ping successfully sometimes in Beijing, if not set the parameter, default is 8.8.4.4.

### DIFF
--- a/WS2012R2/lisa/setupscripts/InjectIPconstants.ps1
+++ b/WS2012R2/lisa/setupscripts/InjectIPconstants.ps1
@@ -138,7 +138,7 @@ $switchs = $False
 $AddressFamily = "IPv4"
 $testMAC = "no"
 # default remote ping server
-$remote_Server = "8.8.8.4"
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/setupscripts/InjectIPconstants.ps1
```diff
@@ -137,7 +137,8 @@ $nic = $False
 $switchs = $False
 $AddressFamily = "IPv4"
 $testMAC = "no"
-
+# default remote ping server
+$remote_Server = "8.8.8.4"
```
chvalean: I think this needs to be 8.8.4.4

lixuemin2016: @chvalean Thank you Chris, have updated.

PR #767, opened 2017-08-31
Minor update STOR_VSS_BackupRestore to make case more stable

lixuemin2016:
Minor update SROR_VSS_BackupRestore_ISO_NoNetwork case
1) after send "at -f runtest.sh now + 1 minutes" by plink, it gets error "Error:Unable to submit runtest.sh to the vm", so ignore the return value checking, it could be covered by pingresult checking.
2) Extend the Start-sleep -Seconds 65 to check the network status
3) WaitforVMToStartKVP failed in Fedora, so update to WaitforVMToStartSSH.

### DIFF
--- a/WS2012R2/lisa/setupscripts/STOR_VSS_BackupRestore_ISO_NoNetwork.ps1
+++ b/WS2012R2/lisa/setupscripts/STOR_VSS_BackupRestore_ISO_NoNetwork.ps1
@@ -24,19 +24,19 @@
     This script tests VSS backup functionality.
 
 .Description
-    This script will stop networking and attach a CD ISO to the vm. 
-    After that it will perform the backup/restore operation. 
-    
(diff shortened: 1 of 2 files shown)


chvalean:
Thank you for the improvements @lixuemin2016 

One question from reviewing the code - I see that you've switched functions to WaitForVMToStartSSH.
This might be a problem if the VM IP changes between the original VM and the time it gets restored, so then the IP would be different - and WaitForVMToStartSSH  relies on the ipv4 param.

lixuemin2016:
Thank you @chvalean , yes, my mistake directly changes WaitForVMToStartKVP to WaitForVMToStartSSH and WaitForVMToStartSSH replies on ipv4 param.

I still could not find the reason why fedora failed after restore vm then start vm, even it works by manual execution and pass on rhel7. I need to further check related log.

Issue #764, opened 2017-08-29
fixes for Core_Config_TimeSync.sh


chvalean:
In [ConfigPtp4l](https://github.com/LIS/lis-test/blob/master/WS2012R2/lisa/remote-scripts/ica/Core_Config_TimeSync.sh#L95)
ptp4l is provided on RHEL and Ubuntu by linuxptp package - must be installed during setup.

Handle different service name in [ConfigChrony](https://github.com/LIS/lis-test/blob/master/WS2012R2/lisa/remote-scripts/ica/Core_Config_TimeSync.sh#L85):
On rhel service name is chrony, ubuntu refers to it as chronyd

PR #771, opened 2017-09-05
Updated TimeSync configuration script

ilenghel:
Updated time sync configuration script to address issue #764

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/CORE_TimeSync_NTP.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/CORE_TimeSync_NTP.sh
@@ -163,6 +163,7 @@ elif is_ubuntu ; then
     service ntp restart
     if [[ $? -ne 0 ]]; then
         LogMsg "NTP is not installed. Trying to install..."
+        apt-get update
         apt-get install ntp -y
         if [[ $? -ne 0 ]] ; then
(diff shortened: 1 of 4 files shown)


chvalean:
@lixuemin2016 can you please also verify these changes and let us know if you have any feedback.
Thank you!

lixuemin2016:
@chvalean @ilenghel  we are checking these changes, will update the feedback in recent two days. Thank you so much.

ilenghel:
Thank you for the feedback. Regarding "return $ptp", it shall be removed. LinuxPTP provides ptp4l and that is why we need to install it. About moving some of the lines to separate functions, because we have to handle different distributions, some of the lines will still repeat. Thank you.

lixuemin2016:
Thank you, base on comment from developer,  he suggests that we no need to cover linuxptp/ptp4l here -  "these are for 'real' PTP devices and their configuration and we're just piggy-backing on PTP protocol." Could you please set the install linuxptp/ptp4l as optional, or not add this part to RHEL case? Thank you so much.

chvalean:
Hi Li,
The ptp4l appeared in the initial commit with the chronny off new tests.
If we don't rely on linuxptp/ptp4l, then we need to remove all the installation and configuration for that.

lixuemin2016:
@chvalean, agree to remove all the installation and configuration for ptp4l based on developer Vitaly's comment.
Thank you so much.

PR #807, opened 2017-10-20
Update DM related script to remove vm definition in case xml parameter

lixuemin2016:
Orignial DM.xml includes VMName, which is used to define the test vmName and dependency vmName. Since two vms name can be defined in the VMs section, we could use and avoid defining them in the testParams section.

Also we run all the cases in one xml file, vmName and VM2Name are defined dynamically, it cannot change the DM.xml single case's parameter by automation, so submit this pull requests.

Thank you.

### DIFF
--- a/WS2012R2/lisa/xml/DM_Tests.xml
+++ b/WS2012R2/lisa/xml/DM_Tests.xml
@@ -22,13 +22,13 @@
 				<suiteTest>DM_loaded</suiteTest>
 				<suiteTest>HotAdd_Verify_udev</suiteTest>
 
-				<!--Dynamic Memory – Hot Add--> 
+				<!--Dynamic Memory – Hot Add-->
 				<!-- Basic Hot Add tests -->
 				<suiteTest>HotAdd</suiteTest>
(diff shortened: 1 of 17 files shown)


WS2012R2/lisa/xml/DM_Tests.xml
```diff
@@ -181,21 +174,19 @@
...
 				<param>staticMem=1024MB</param>
-				<param>vmName=VM2Name</param>
 				<param>enableDM=yes</param>
 				<param>minMem=1024MB</param>
 				<param>maxMem=40%</param>
 				<param>startupMem=40%</param>
 				<param>memWeight=0</param>
 				<param>staticMem=1024MB</param>
-				<param>vmName=VM3Name</param>
+				<param>vm3Name=VM3Name</param>
```
chvalean: I see that your changes are addressed in the VM section of the xml, however vm3 is still kept here.
Would it be the same to move vm3 in the xml VM section to keep the same new format proposed?
Thank you!

lixuemin2016:
Chris, thank you for your comment, have updated.

Thank you.

PR #813, opened 2017-10-26
Update TCUtils with install_app for DM related stress-ng and stressap…

lixuemin2016:
…ptest

Add function install_app to TCUtils.ps1 and update DM related script from check_app to install_app.

Notice that already there is one file aio.sh to prepare test environment and install stress-ng and stressapptest, locally we do not run aio.sh, so try this resolution to install stress-ng and stressapptest.

Thank you.

### DIFF
--- a/WS2012R2/lisa/setupscripts/TCUtils.ps1
+++ b/WS2012R2/lisa/setupscripts/TCUtils.ps1
@@ -1301,15 +1301,15 @@ function check_kernel
 # Check for application on VM
 #
 #######################################################################
-function check_app([string]$app_name, [string]$custom_ip)
+function checkApp([string]$appName, [string]$customIP)
 {
-    IF([string]::IsNullOrWhiteSpace($custom_ip)) {
-        $target_ip = $ipv4
+    IF([string]::IsNullOrWhiteSpace($customIP)) {
+        $targetIP = $ipv4
     }
     else {
-        $target_ip = $custom_ip
+        $targetIP = $customIP
     }
-    .\bin\plink -i ssh\${sshKey} root@${target_ip} "command -v ${app_name}"
+    .\bin\plink -i ssh\${sshKey} root@${targetIP} "command -v ${appName} > /dev/null 2>&1"
     if (-not $?) {
         return $False
     }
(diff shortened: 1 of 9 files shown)


WS2012R2/lisa/setupscripts/TCUtils.ps1
```diff
@@ -1311,11 +1311,62 @@ function check_app([string]$app_name, [string]$custom_ip)
...
+        $app_version="V0.07.16"
+    }
+    elseif ($app_name -eq "stressapptest")
+    {
+        $app_githubLink="https://github.com/stressapptest/stressapptest.git"
+    }
+
+    .\bin\plink -i ssh\${sshKey} root@${target_ip} "cd /root; git clone $app_githubLink $app_name"
+
+    if ($app_version)
```
chvalean: would it be possible to script the installation of the tools independently, similar to the aap_name if loop?
This was we can re-use the install_app function later on

lixuemin2016: @chvalean thank you for your comment.

I think I can script the installation of the tools, but need some guidance about how to implement it, I'm not sure the requirement. Could you please help to explain more?

1) "For independently", do you mean "install_app" function itself, I need to remove check_app or put it in other place? 
2) I'm not very clear about "similar to the app_name if loop", you mean support app name as list, then can install multiple app at the same time, such as put the app_name and github link, github version to dictionary, then the list includes dictionary?  Currently if app_name is not stress-ng or stressapptest, it also can support, but when provide app name, needs to provide git link.

Thank you so much.

chvalean: I was thinking at having install_app in a more universal way, so that in the future we can extend it to install other tools as well.
Currently with the new changes, it is more or less hardcoded to handle only stress-ng and stressapptest with their specific method of installation.

One suggestion would be to encapsulate any tool installation in the name parsed - eg if ($app_name -eq "stress-ng")

lixuemin2016:
Chris, thank you for your comment.
Understand, before think if not hard-code here, it needs to get the app name, app git url and tag version by xml, then it needs to change more DMTest.xml and DM script parse parameter, it is not good idea to hard code here.
If in a more universal way, for DM, we need to add these app information, git information in the test case xml definition, do you agree that?

I will submit new version later.

chvalean:
That would be even better, including URLs and tag versions inside some script is not efficient in case we have to update that long term, so the xml would be more appropriate.

lixuemin2016:
@Chris, could you please help to review this pull request about installApp? If any comment, I will update.
Thank you so much.

PR #827, opened 2017-11-10
Update stateEngine runPostTestScript to avoid exit test when fail

lixuemin2016:
remove 3 lines, set currentTest="done" and updateState as finisheded.

Currently issue is that if postTestScript failed, it will exit the whole Lisa test.

Thank you


chvalean:
Thanks for the improvement @lixuemin2016 
Quick question - do we might want to keep $vm.currentTest = "done"?
Or lisa will move to the next test anyhow?

lixuemin2016:
Thank you Chris.
If postTestScript failed, test script passed, currently treat the test case is passed, only output the error log.
Lisa will move to the next test, it will got to "UpdateState $vm $DetermineReboot" line.
We have one pull code coverage script as postscript for every case, every time if this postscript is failed, then it directly exits LISA, it seems not reasonable.

PR #835, opened 2017-11-28
added install kernel-devel if not exist

ampintilie:


WS2012R2/lisa/remote-scripts/ica/install_lis_next.sh
```diff
@@ -176,6 +175,17 @@ cd "${lis_next_path}/hv-rhel${rhel_version}.x/hv"
 # Defining a custom LIS version string in order to acknoledge the use of these drivers
 sed --in-place -e s:"#define HV_DRV_VERSION.*":"#define HV_DRV_VERSION "'"'$branch'-'$build_date'"'"": include/linux/hv_compat.h
 
+
+if ! [[ $(rpm -qa | grep kernel-devel-$(uname -r)) ]]; then
+    yum install kernel-devel-$(uname -r) -y
+    cd /lib/modules/$(uname -r)/
```
chvalean: this is not required, installing the package will be enough.

Issue #840, opened 2017-12-08
Automatically collect console output in lisa


chvalean:
We need to implemented a method that will configure the COM2 port for new VMs and capture the VM serial log during the entire duration of testing.

PR #841, opened 2017-12-13
 Save VM output to log file

bogdancarpusor:
A new feature was added in order to save the output
 of the tested VMs in separate log files.
 Output log files are created for each VM used and
 saved in the test suite log path, under the following
 name format "${vmName}-icaserial.log"

 Also the CreateVMs setup script was update in order
 to configure the com2 port with an unique path.


chvalean:
Feature request https://github.com/LIS/lis-test/issues/840

PR #842, opened 2017-12-18
remote-scripts/ica/Hyperv_Daemons_Files_Status.sh: No need to check t…

HuijingHei:
No need to check the hyperv daemons status in /lib/systemd/system-preset/90-default.preset.

From Vitaly according to https://bugzilla.redhat.com/show_bug.cgi?id=1513554#c4:
hypervfcopyd.service (+hypervvssd.service, hypervkvpd.service) don't require manual enablement at all. These services are started automatically when corresponding devices appear.

### DIFF
--- a/Tools/AzureSubUtilizationReport/AddAzureRmAccountFromSecretsFile.ps1
+++ b/Tools/AzureSubUtilizationReport/AddAzureRmAccountFromSecretsFile.ps1
@@ -1,0 +1,74 @@
+﻿<#
+.SYNOPSIS
+    This script authenticates PS sessing using Azure principal account.
+
+.DESCRIPTION
+    This script authenticates PS sessing using Azure principal account.
+
+.PARAMETER -customSecretsFilePath
+    Type: string
+    Required: Optinal.
+
+.INPUTS
+    AzureSecrets.xml file. If you are running this script in Jenkins, then make sure to add a secret file with ID: Azure_Secrets_File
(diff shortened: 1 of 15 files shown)


chvalean:
Would it be good to know if this remains applicable to old RHEL 6.x and 7.x releases.
I didn't verify these yet, can you please get a confirmation on that?
Thank you!

HuijingHei:
@chvalean  Thanks for your review!

1) /lib/systemd/system-preset/90-default.preset is only for rhel7.x, I have checked rhel7.4 and rhel7.3, these hyperv services are started automatically when corresponding devices appear.
$ cat hypervfcopyd.service 
[Unit]
Description=Hyper-V FCOPY daemon
BindsTo=sys-devices-virtual-misc-vmbus\x21hv_fcopy.device

[Service]
ExecStart=/usr/sbin/hypervfcopyd -n

2) For old rhel7.2, the hypervfcopyd.service (+hypervvssd.service, hypervkvpd.service) depend on ConditionVirtualization=microsoft, perhaps still need enable in /lib/systemd/system-preset/90-default.preset. But I don't think we should cover the old rhel7.2 version.

chvalean:
At least for LIS 4.x we must provide compatibility for older releases as well.
If this new pattern is valid for 7.3 and newer, can you please add a condition for that maybe?

HuijingHei:
OK, I will commit the new patch according to your suggestion later. Thanks!

PR #843, opened 2017-12-18
SR-IOV: Added 'One vCPU' TC

adriansuhov:
Due to recent failures, a TC that tests SR-IOV with only 1 vCPU is needed

### DIFF
--- a/WS2012R2/lisa/xml/SR-IOV.xml
+++ b/WS2012R2/lisa/xml/SR-IOV.xml
@@ -65,6 +65,7 @@
                 <suiteTest>Measure_DisableVF</suiteTest>
                 <suiteTest>Move_VHD</suiteTest>
                 <suiteTest>Max_vCPU</suiteTest>
+                <suiteTest>One_vCPU</suiteTest>
                 
                 <!-- Mellanox specific -->
                 <suiteTest>Upgrade_Downgrade_PF</suiteTest>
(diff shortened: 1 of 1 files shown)


WS2012R2/lisa/xml/SR-IOV.xml
```diff
@@ -391,6 +392,26 @@
...
+            <testScript>SR-IOV_VerifyVF_basic.sh</testScript>
+            <files>remote-scripts/ica/SR-IOV_VerifyVF_basic.sh,remote-scripts/ica/utils.sh,remote-scripts/ica/SR-IOV_Utils.sh</files> 
+            <setupScript>
+                <file>setupscripts\RevertSnapshot.ps1</file>
+                <file>setupScripts\ChangeCPU.ps1</file>
+                <file>setupscripts\SR-IOV_enable.ps1</file>
+            </setupScript> 
+            <noReboot>False</noReboot>
+            <testParams>
+                <param>NIC=NetworkAdapter,External,SRIOV,001600112800</param>
```
chvalean: do we require to use static MAC?

WS2012R2/lisa/xml/SR-IOV.xml
```diff
@@ -391,6 +392,26 @@
...
+                <param>TC_COVERED=SRIOV-22a</param>
+                <param>vCPU=1</param>
+                <!-- This is based on the physical CPU hardware topology -->
+                <param>numaNodes=36</param>
```
chvalean: this should not be required

adriansuhov: Fixed

WS2012R2/lisa/xml/SR-IOV.xml
adriansuhov: Removing the MAC address would require additional code change & extended testing. I suggest we note it in the Issues tab as an automation improvement to do and work on it separately.

PR #845, opened 2017-12-19
modules_check.sh: Remove skipped gen1

HuijingHei:
Remove skipped gen1, including regression test for gen1


chvalean:
Hello,
This script handles the LIS modules validation in the rescue boot image itself.
For Gen2 VMs only we are required to have the keyboard driver for eg, but for gen1 we don't have any such mandatory requirements.
Therefore we skip the check on gen1.

Please let me know if you have other info on this.

PR #862, opened 2018-01-11
Improved daemon name handling

ilenghel:
Improved the way in which the daemon name is being handled by the
test script for better accuracy.

### DIFF
--- a/WS2012R2/lisa/setupscripts/FCOPY_disableEnable.ps1
+++ b/WS2012R2/lisa/setupscripts/FCOPY_disableEnable.ps1
@@ -305,10 +305,14 @@ if ($checkVM -eq "True") {
         return $Failed
     }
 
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/setupscripts/FCOPY_disableEnable.ps1
```diff
@@ -305,10 +305,14 @@ if ($checkVM -eq "True") {
         return $Failed
     }
 
-    $checkProcess = .\bin\plink.exe -i ssh\${sshKey} root@${ipv4} "systemctl is-active *fcopy*"
+    # Daemon name might vary. Get the correct daemon name based on systemctl output
```
chvalean: recommended to make us of FCOPY_Check_Daemon.sh to handle the daemon check

PR #867, opened 2018-01-16
Update storage resize limiation for host from 50G to 20G

lixuemin2016:
Minor updates:
1) update host storage size flag from 50GB to 20GB, since our test host has less storage, if the limitation set  as 50GB, storage resize related test cases will fail. Have check resize related test cases, generally it will change from 3G size to 4G. For 2304GB, if it is format as ext4, disk size is about 1.1G, if set as ext3, disk size will be 150G, here ignore ext3 file system.

In the file STOR_VHDXResize_PartitionDiskOver2TB.sh line 129, if it could mkfs successfully, will break the look, so it only test ext4 here. 

2)  Remove the ext3 from the xml when resize more than 2T.

3) Minor update logger.info from logger.error for "Restore duration" information for backup test case.


chvalean:
Thank you @lixuemin2016  for the changes.
I do agree that as we cover ext4, it is safe to remove ext3 at this point.
The 50GB limit might have been set due to ext3 higher disk requirements for format on larger disks.

PR #871, opened 2018-01-18
Add param "Large_File_Size_GB" for FCOPY_large_file case

zqwzshm:
This patch is to allow users customize file size in xml file.

If this param is not set in xml, file size will still be 10GB by default.
Thus no modification needed for current xml files.


chvalean:
Can you please add the Large_File_Size_GB param in the FCopy xml as well, maybe commented out with a description of it?
That way it will be understood that it can be customized.
Thank you!

PR #872, opened 2018-01-18
Add param "Large_File_Size_GB" for FCOPY_large_file case

zqwzshm:
If the param not set in xml, file size will be 10GB by default.


zqwzshm:
added commented sample param settings as you suggested, thanks

PR #895, opened 2018-01-30
Add UIO test case - UIO_Basics

zqwzshm:
This test covers basic operations of uio and uio_hv_generic:
  1. loading/unloading uio & uio_generic for 100 times
  2. assign a device (network adapter) to uio_hv_generic

Manal test steps:
1. setup another network adapter for the vm
2. Load uio, check it is loaded
  \# modprobe uio
  \# lsmod |grep uio
3. Unload uio, check it is unloaded
  \# modprobe -r uio
  \# lsmod |grep uio
4. Load uio_hv_generic, check uio & uio_hv_generic are loaded
  \# modprobe uio_hv_generic
  \# lsmod |grep uio
5. Add new id (ClassID) to uio_hv_generic
  \# ip addr
  ( Use new network adapter for the ehtx in following commands )
  \# ClassID=`cat /sys/class/net/ethx/device/class_id |awk '{ print( substr($1, 2, length($1)-2)) }'`
  \# echo $ClassID > /sys/bus/vmbus/drivers/uio_hv_generic/new_id
6. Deassign the device from kernel hv_netvsc driver
  \# DeviceID=`cat /sys/class/net/ethx/device/device_id |awk '{ print( substr($1, 2, length($1)-2)) }'`
  \# echo -n $DeviceID > /sys/bus/vmbus/drivers/hv_netvsc/unbind
  (Make sure to use 'echo -n')
7. Assign the device to uio_hv_generic
  \# echo -n $DeviceID > /sys/bus/vmbus/drivers/uio_hv_generic/bind
  Check /dev/uio0 exists.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/UIO_Basic.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/UIO_Basic.sh
@@ -1,0 +1,212 @@
+#!/bin/bash
+
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the ""License"");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+# OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
+# ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR
+# PURPOSE, MERCHANTABLITY OR NON-INFRINGEMENT.
+#
+# See the Apache Version 2.0 License for specific language governing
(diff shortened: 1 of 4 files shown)


zqwzshm:
It is a new feature we have been testing. We created this test following this bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1396534

I'm not sure if I should make this test RHEL only, so have not add distro_check part yet.
We currently use an inactive network adapter to assign to uio, I am not sure if we need to use a connected network adapter instead. Would it do any help?

WS2012R2/lisa/remote-scripts/ica/UIO_Basic.sh
```diff
@@ -0,0 +1,212 @@
...
+	echo "TestAborted" > state.txt
+	exit 2
+}
+
+#######################################################################
+# Check kernel version is newer than the specified version
+# if return 0, the current kernel version is newer than specified version
+# else, the current kernel version is older than specified version
+#######################################################################
+CheckVMFeatureSupportStatus()
```
chvalean: can this be moved to utils.sh and be sourced from there?
This function already exists on the PS side in tcutils.ps1, so it might be more useful for future uses in the utils functions.
Thank you!

zqwzshm:
Modified according to your suggestion
Thanks a lot for reviewing!

chvalean:
Thank you for submitting the test case!

PR #934, opened 2018-03-08
Fix: TCUtils.ps1: GetVMFeatureSupportStatus wrong return value

zqwzshm:
The patch 1406f1b #915 is bringing back a issue fixed in patch 16d1884 #905 .
eg. Feature is supported in version "3.3.3-3", current kernel version is "3.2.4-3", function would return $True, which is incorrect.

### DIFF
--- a/WS2012R2/lisa/setupscripts/KVP_TestRescind.ps1
+++ b/WS2012R2/lisa/setupscripts/KVP_TestRescind.ps1
@@ -158,8 +158,8 @@ if ($checkVM -eq "True") {
         return $Failed
     }
 
-# Check if VM is RedHat 7.2 or older and if it uses external LIS
-    $supportkernel = "3.10.0.327" #kernel version for RHEL 7.2
+    # Check if VM is RedHat 7.3 or later (and if not, check external LIS exists)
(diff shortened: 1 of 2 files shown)


ilenghel:
Hello,
Commit #915 was meant to fix a problem we had while testing a certain scenario. In order for those changes not to break the KVP_TestRescind scenario, could you please also update the  $supportkernel = "3.10.0.514" in KVP_TestRescind.ps1?

zqwzshm:
Hi,
I have updated the kernel version to 3.10.0.514, thanks for reviewing!

But the log message is confusing me somehow. I think the the kernel version checking part in KVP_TestRescind is to check rhel kernel >= 3.10.0.514, and if not, check external LIS exists. (Correct me if I am wrong). While the log message in the code is "Info: Kernels older than 3.10.0-514 require LIS-4.x drivers.". I am not sure what it actually means. Is it trying to say "kernel version before 514 requires LIS external"?

So, I am not sure if I just updated it in the right way. Would you please help check it again, thanks!

zqwzshm:
ok, thanks for the explanation, now it make sense to me. :-)

PR #936, opened 2018-03-09
Add hv-sock test case: HV_Sock_Basic

zqwzshm:
This test covers basic socket connection and data transmission using
hv-sock.

Part I: Client app on vm connects to server app on host
Part II: Client app on host connects to server app on vm

Manual test steps:

Setup:

1. Add register key for GuestCommunicationService on host:

    On host, run "regedit", find open the path:
    HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\
    Virtualization\GuestCommunicationServices\

    If the key "00000808-facb-11e6-bd58-64006a7986d3" does not exists,
    add it, set value "HV Socket Service" for the key

2. Load module on guest linux:
    \# modprobe hv-sock

Test Part I:

1. Run server_on_host.exe on host
2. Run client_on_vm on guest linux
3. Check output on both sides, connected successfully without error.

Test Part II:

1. Get VM GUID in powershell:
    > Get-VM -Name "Your VM Name" |fl name, id
2. Run server_on_vm on guest linux
3. Run client_on_host.exe on host with argument VM GUID:
    > client_on_host.exe
3. Check output on both sides, connected successfully without error.


WS2012R2/lisa/xml/HV_Sock.xml
```diff
@@ -0,0 +1,70 @@
...
+    ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR
+    PURPOSE, MERCHANTABLITY OR NON-INFRINGEMENT.
+
+    See the Apache Version 2.0 License for specific language governing
+    permissions and limitations under the License.
+-->
+
+<config>
+
+    <global>
```
chvalean: See the dependency tag here - https://github.com/LIS/lis-test/blob/master/WS2012R2/lisa/xml/SR-IOV.xml#L26

You can add that code to have a lisa higher level validation that if the host does not meet the minimum version requirement, it will exit before running any test code.

chvalean:
Thank you for submitting the new test case @zqwzshm !
Unfortunately I won't be able to accept the 2 windows executables as is.
Will have to compile them on our after publish, please let me know if there are any special compile instructions.

zqwzshm:
Modified according to your sugestions
Compiling method is given in tools\hv-sock\README file
Thank you!

ilenghel:
Hello,
I have tested these implementations, but I didn't manage to get a passing result. Regarding my environment: the hv_sock module is loaded on the VM and other applications that use the feature are working as expected. The registry has been modified accordingly and the host platform supports the feature. I have compiled the source codes according to the README file. I have tested both automated run and manual run, and neither one worked. Are there any other requirements that I should be aware of?
Thank you!

zqwzshm:
Hi,
I downloaded the server_on_host.cpp directly from this branch and created a empty c++ vs project then built it, everything works fine for me. I ran the generated exe on 2 hyper-v server core hosts, nothing went wrong.

I am a little puzzled why this is happening. The server_on_host does not have much dependencies. You were able to run the executable and even get some output, which means the executable were able to be loaded and dependencies exists. The server_on_host.exe is with least dependencies, it does not even require response from vm to run.

I'll send you the original vs project provided by Dexuan Cui via email slightly later. Let's see if it works out.

dcui:
@ilenghel 
> socket() failed: error = 10022 
Please see https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668(v=vs.85).aspx, it means: WSAEINVAL

Are you running the program on WS 2012 R2? hv_sock is only supported on WS 2016 (Win10).

If you're indeed using 2016, please share your host and I can RDP into it and have a look.

ilenghel:
Hello,
The problem seems to have been caused by my default settings for C++ projects (Win32 platform instead of x64). While for the original applications I had the entire project which kept its platform settings, in this case I started from a new empty project and the default settings from VS were applied.
With the new executables built, I re-run the automation for the hv_sock feature and it was Passed.

Thank you!

zqwzshm:
Sorry about forgetting that part in readme file, updated accordingly. Thank you!

PR #947, opened 2018-03-15
 KVP_Max_Read: skip kernel version below 3.10.3.860 (RHEL7.5)

zqwzshm:
KVP_Max_Read: skip test if kernel version is below 3.10.3.860 (RHEL7.5)
This version is not official release version yet, might need to update later

Some modification in function GetVMFeatureSupportStatus, makes it support both format like "10.3.0.1" and format like"10.3.0-1"


chvalean:
Thank you for the addition here.
Unfortunately this change does not cover a scenario where LIS 4.x is installed, which provides the fix.

zqwzshm:
Modified, thanks for reviewing

PR #948, opened 2018-03-15
Minor update kdumpconfig

lixuemin2016:


WS2012R2/lisa/remote-scripts/ica/Kdump_Config.sh
```diff
@@ -138,7 +138,7 @@ ConfigRhel()
         GetOSVersion
     fi
 
-    if [[ $os_RELEASE.$os_UPDATE =~ ^5.* ]] || [[ $os_RELEASE.$os_UPDATE =~ ^6.[0-2] ]] ; then
+    if [[ $os_RELEASE.$os_UPDATE =~ ^5.* ]] || [[ $os_RELEASE.$os_UPDATE =~ ^6.[0-2][^0-9] ]] ; then
```
chvalean: The additional kdump settings are required only up to RHEL 6.2.
In what configuration are these applicable to all 6.x series?

lixuemin2016: Hi Chirs, thank you for your comment. 
Here mainly consider that RHEL6.10, when testing on RHEL6.10, it match 6.[0-2],then do RHELExtractingSettings, in the RHELextractingSettings, it has blacklist, then after trigger kdump, cannot find vmcore under /var/crash. 

Thank you so much.

PR #956, opened 2018-03-23
Kdump: support testing RHEL6/7 on WS2012

zqwzshm:
Extra kdump config is required on WS2012, while previous extra config
does not work for WS2012 host.

Test-NetConnection does not work for WS2012, use function
WaitForVMToStartSSH in TCUtils instead.

### DIFF
--- a/WS2012R2/lisa/setupscripts/Kdump.ps1
+++ b/WS2012R2/lisa/setupscripts/Kdump.ps1
@@ -18,7 +18,6 @@
 # permissions and limitations under the License.
 #
 #####################################################################
-
 param([string] $vmName, [string] $hvServer, [string] $testParams)
 
 $retVal = $false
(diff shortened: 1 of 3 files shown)


zqwzshm:
This patch might still have some problem, I need more time to double check, closing this PR.
Sorry for the trouble

zqwzshm:
The kdump cases are not stable, refetch ipv4 of VM after config reboot and crash reboot in case a new ipv4 allocated to the VM.

We have a known issue that kvp would not work after kdump on WS2016 - RHEL6. This bug has been confirmed won't be fixed, so I had to reboot the VM again after kdump so that new ip address could be retrieved by kvp.
https://bugzilla.redhat.com/show_bug.cgi?id=1383037

WS2012R2/lisa/setupscripts/Kdump.ps1
```diff
@@ -257,35 +254,24 @@ else {
...
-        Stop-VM -Name $vmName -ComputerName $hvServer -Force
-        return $false
-    }
+Start-Sleep 10
+
+# Workaround a won't be fixed bug on WS2016 - RHEL6:
+# https://bugzilla.redhat.com/show_bug.cgi?id=1383037
+if ( (-not $RHEL7_Above) -and ($BuildNumber -eq "14393") ){
+    Start-Sleep 120 # Make sure dump completed
+    Stop-VM -vmName $vmName -ComputerName $hvServer -TurnOff
```
chvalean: might I suggest a -Force param here as well?
If we VM is in a blocked state after 120 seconds, the TurnOff command will fail.

zqwzshm:
Fixed, thank you!
Also fixed a stability issue that crash on multiple core would abort with timeout sometimes.

chvalean:
Thank you for the improvements to the code!

PR #977, opened 2018-04-11
STOR_Large_Disk.sh: skip IDE test for Gen2 VM

zqwzshm:


chvalean:
Thank you for the code improvement here.
One suggestion I might have is that instead of adding a new variable, it might be more direct to simply use diskCount var as it is now, and after the ide* if check [(here)](https://github.com/LIS/lis-test/blob/master/WS2012R2/lisa/remote-scripts/ica/STOR_Large_Disk.sh#L216) 
to add there directly the check from your PR for matching diskCount and efi presence.

zqwzshm:
modified, thank you for your suggestion.
the code was from [another script](https://github.com/LIS/lis-test/blob/84de4130c9004bb64ff1e9ca5cde28a3d867f44a/WS2012R2/lisa/remote-scripts/ica/STOR_Lis_Disk.sh#L234) which is working correctly, so I just copied the code snip.

PR #991, opened 2018-04-19
stateEngine: Fix two incorrect xml references

zqwzshm:
When starting a preTest script, console output would be like
"starting preTest script System.XML.XmlElement"

If VM become unable to start after running setupScript, test case
name in test result will be set to "System.XML.XmlElement"


zqwzshm:
found another incorrect reference, working on it,
closing this pr for now, will return with another patch for that

zqwzshm:
One more patch for another similar issue

PR #1016, opened 2018-05-09
Minor update HV_SOCK_Basic Test_Type since local have different TC_Co…

lixuemin2016:
…vered value

Locally have random test case ID (without 1,2 at the end of name), cannot use TC_COVERED to select test part, so add TEST_TYPE here.


chvalean:
Thank you for the changes!
We were looking to actually not add an additional param just for this, but if you're using different test cases IDs then it makes sense.

PR #1042, opened 2018-06-07
Add clockevent check case

HuijingHei:
Not support on rhel6

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/check_clockevent.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/check_clockevent.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the ""License"");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+# OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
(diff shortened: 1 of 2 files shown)


WS2012R2/lisa/remote-scripts/ica/check_clockevent.sh
```diff
@@ -0,0 +1,95 @@
...
+    fi
+
+}
+
+#
+# MAIN SCRIPT
+#
+GetDistro
+case $DISTRO in
+    redhat_6)
```
chvalean: it should also contain centos_6

WS2012R2/lisa/remote-scripts/ica/check_clockevent.sh
```diff
@@ -0,0 +1,95 @@
...
+        LogMsg "WARNING: $DISTRO does not support clockevent."
+        UpdateSummary "WARNING: $DISTRO does not support clockevent."
+        SetTestStateSkipped
+        exit 1
+        ;;
+    redhat_7 | redhat_8)
```
chvalean: please also add centos_7 and centos_8

WS2012R2/lisa/remote-scripts/ica/check_clockevent.sh
```diff
@@ -0,0 +1,95 @@
...
+        CheckClockEvent
+        ;;
+    fedora_* )
+        CheckClockEvent
+        ;;
+    *)
```
chvalean: we usually also support ubuntu, if it has this feature then we should handle it

WS2012R2/lisa/xml/CoreTests.xml
```diff
@@ -382,6 +383,19 @@
             <noReboot>True</noReboot>
         </test>
 
+        <test>
+            <testName>check_clockevent</testName>
+            <testScript>check_clockevent.sh</testScript>
+            <files>remote-scripts/ica/check_clockevent.sh</files>
```
chvalean: multiple files can be handled in a single definition by using coma

HuijingHei:
Hi Chris, I have updated scripts according to your comments. Thanks!

PR #1056, opened 2018-06-20
Add clockevent check on SMP CPUs

HuijingHei:
1. Only check clocksource on rhel6, skip unbind for rhel6
2. check and unbind clockevent on UP for rhel7, skip on rhel6
3. check and unbind clockevent on SMP for rhel7, skip on rhel6


HuijingHei:
1. Update check_clocksource to add unbind clocksource function, rename the suitname and filename
2. Update UP CPU to add unbind clockevent function, rename the suitname and filename
3. Update SMP CPU to skip unbind clockevent check

PR #1084, opened 2018-07-19
Add one new hypervfcopyd test case for negative test

lixuemin2016:
Add new test cases to do negative test for hyeprvfcopyd, verify that fails to copy file from Windows host to guest for following three scenarios .
1) when target folder is immutable
2) 'Guest Service Interface' is disabled/unchecked in the vm settings.
3) hyperverfcopyd is disabled. (restart hypervfcopyd again to make sure no file copied to vm)

Note that on the latest rhel7, disable-vmintegrationservice can turn on and take effect directly. It no needs to stop vm then restart vm. Before maybe not work for rhel6. During testing, checks systemd available or not, it will skip rhel6. 

Thank you so much.


WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
```diff
@@ -0,0 +1,167 @@
...
+
+    return $True
+}
+
+#################################################################
+#
+# Check systemd available or not
+#
+#################################################################
+function Check-Systemd()
```
chvalean: let's get this function under TCUtils.ps1. it already exists into 2 other separate scripts, so we can consolidate it under tcutils.

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
```diff
@@ -0,0 +1,167 @@
...
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+# OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
+# ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR
+# PURPOSE, MERCHANTABLITY OR NON-INFRINGEMENT.
+#
+# See the Apache Version 2.0 License for specific language governing
+# permissions and limitations under the License.
+#
+########################################################################
+
```
chvalean: all these functions should be moved to tcutils.ps1.
More complex test areas do have a separate utils script, but for fcopy we should have functions generic enough to be able to put them under tcutils.ps1.

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
```diff
@@ -0,0 +1,167 @@
...
+#######################################################################
+#
+#	Checks if the file copy daemon is running on the Linux guest
+#
+#######################################################################
+function check_fcopy_daemon()
```
chvalean: with this under tcutils.ps1, we can remove each duplicate function from all fcopy scripts

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
```diff
@@ -0,0 +1,167 @@
...
+    }
+    return $true
+}
+
+#######################################################################
+#
+#	Checks if test file is present
+#
+#######################################################################
+function check_file([String] $testfile)
```
chvalean: same for this

WS2012R2/lisa/setupscripts/FCOPY_negative.ps1
```diff
@@ -0,0 +1,271 @@
...
+    return $Skipped
+}
+
+# Source FCOPY_Utils.ps1
+. .\setupScripts\FCOPY_utils.ps1
+
+# If vm does not support systemd, skip test.
+$sts = Check-Systemd
+if ($sts[-1] -eq $false){
+	return $Skipped
```
chvalean: can you please add a logged message here that "Distro does not support systemd, skipping test."?
That way we know directly the reason when reviewing the results.

lixuemin2016:
Hi Chris, thank you so much for your review comment, will update it.

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
lixuemin2016: Done, thank you.

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
lixuemin2016: Done, thank you.

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
lixuemin2016: Done, thank you.

WS2012R2/lisa/setupscripts/FCOPY_utils.ps1
lixuemin2016: Done, thank you.

WS2012R2/lisa/setupscripts/FCOPY_negative.ps1
lixuemin2016: Done

WS2012R2/lisa/setupscripts/fcopy_repeated_delete.ps1
```diff
@@ -58,102 +58,19 @@ param([string] $vmName, [string] $hvServer, [string] $testParams)
...
-    if (-not $sts) {
-        Write-Output "ERROR: Failed to make file system in the VM $vmName." | Tee-Object -Append -file $summaryLog
-        return $False
-    }
-
-    $sts = SendCommandToVM $ipv4 $sshKey "mount ${driveName}1 /mnt"
-    if (-not $sts) {
-        Write-Output "ERROR: Failed to mount the disk in the VM $vmName." | Tee-Object -Append -file $summaryLog
+####################################################################################
+function check_file_vm(){
```
chvalean: this check_file_vm function should no longer be required, why not just call out directly CheckFile instead?

lixuemin2016: Hi Chirs,
Thank you for your review, will update it to make script more concise. Yes, it can be removed by using checkFile directly, not remove it only because I want to do less change to original script.

lixuemin2016: Done, thank you.

chvalean:
@lixuemin2016  I was testing the new test case and came across a scenario here - my VM did not have the fcopy daemon package installed (7.0 where it was as a separate package for each daemon), however the test case passed successful and reported that all 3 negative scenarios passed.
This of course is a false-positive result and should be addressed.

Also verified the new TC with a VM with fcopy present and the code looks good, so if you can please only cover the above feedback and this is good to go.
PS: also updated the xml for you to have the test case in the actual list.

lixuemin2016:
@chvalean, thank you so much for your view and comment.
Cover your feedback by using check_fcopy_daemon function (similar as other fcopy scripts). Locally I haven't install RHEL7.0 yet, from script's perspective, it should work. If still not work, will update again.

PR #1091, opened 2018-08-10
lsvmbus: fix cpu count issue about network driver and scsi controller

HuijingHei:
1. From my test, the cpu count that attached to the network driver is less than 8, and the cpu count that attached to the scsi controller is (N+3)/4.

Not sure about the cpu count is right, the result is only the same as my test. If it is wrong, feel free to correct me. Thanks!

2. Add fedora support.


chvalean:
Thank you for the fixes!
The changes are looking great, as the netvsc spread caps at 8 cores, we were settings that number of cores in the xml and not pushing beyond, so the fix is on spot in case a higher cpu count is used.

PR #1096, opened 2018-08-20
Minor update CORE_max_cpus.ps1 to fix bug

lixuemin2016:
Minor change: if $guest_max_cpus is larger than host supported locical cores, set $guest_max_cpus as maxCpus.


chvalean:
Thank you for the fix @lixuemin2016 
I recall having that line but probably removed it for some checks. :)

PR #1097, opened 2018-08-20
Update Core_StressReloadModules to deeper stress

lixuemin2016:
Hi Chris,

Currently there is one bug https://bugzilla.redhat.com/show_bug.cgi?id=1613655, if use original #100 times and sleep 1 for six times, it cannot catch the bug like https://bugzilla.redhat.com/show_bug.cgi?id=1613655, this case only reproduces on SMP, but not UP, so may need to add two cases. Do you want to change related cases.xml to add new test cases about StressReloadModules_UP and StressReloadModules_SMP? If yes, I will change it too?

Thank you so much.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/CORE_StressReloadModules.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/CORE_StressReloadModules.sh
@@ -146,22 +146,35 @@ if [ $? -eq 0 ]; then
     exit 1
 fi
 
+# if not set the LoopCount, set as 100 by default
+if [ "${LoopCount:-UNDEFINED}" = "UNDEFINED" ]; then
+    LoopCount=100
+fi
+
+# if not set the sleep duration, set as 1 second by default
+if [ "${Duration:-UNDEFINED}" = "UNDEFINED" ]; then
+    Duration=1
+fi
+msg="Info: module unload/load loop count set as $LoopCount"
+LogMsg "${msg}"
+echo "$msg" >> ~/summary.log
+
 pass=0
 START=$(date +%s)
-while [ $pass -lt 100 ]
+while [ $pass -lt $LoopCount ]
 do
     modprobe -r hv_netvsc
-    sleep 1
+    sleep $Duration
     modprobe hv_netvsc
-    sleep 1
+    sleep $Duration
     modprobe -r hv_utils
-    sleep 1
+    sleep $Duration
     modprobe hv_utils
-    sleep 1
+    sleep $Duration
     modprobe -r hid_hyperv
-    sleep 1
+    sleep $Duration
     modprobe hid_hyperv
-    sleep 1
+    sleep $Duration
     pass=$((pass+1))
     echo $pass
 done
--- a/WS2012R2/lisa/remote-scripts/ica/CORE_StressReloadModules.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/CORE_StressReloadModules.sh
@@ -24,6 +24,7 @@
 ########################################################################
 #
 # CORE_StressReloadModules.sh
+#
 # Description:
 #    This script will first check the existence of Hyper-V kernel modules.
 #    Then it will reload the modules in a loop in order to stress the system.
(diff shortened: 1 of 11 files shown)


chvalean:
Yes @lixuemin2016 , I think that 2 test cases for UP and SMP would be the right way, especially given the origin of the ticket with that behavior.
If we bump the loop count from 100 to 2000, what is the total running time of the test case?
If it's over 30 minutes, I would put it though in the stress test category, or maybe split them into Core test with a limited number of repeats, and a more in-depth check with 2000 iterations that would go into the Stress tests area?

lixuemin2016:
@chvalean , updated case.xml to add two cases StressReloadModules_UP and StressReloadModules_SMP.  After change to 2000 and remove 6 seconds sleep, running time about 4 minutes for one case. It should be ok to keep them in the Core test. 
Thank you so much.

WS2012R2/lisa/remote-scripts/ica/CORE_StressReloadModules.sh
```diff
@@ -146,22 +146,24 @@ if [ $? -eq 0 ]; then
...
+LogMsg "${msg}"
+echo "$msg" >> ~/summary.log
+
 pass=0
 START=$(date +%s)
-while [ $pass -lt 100 ]
+while [ $pass -lt $LoopCount ]
 do
     modprobe -r hv_netvsc
-    sleep 1
```
chvalean: we should keep the sleep commands in here, in the past it was discussed that a short delay must exist in the unload/load of the module. it should also not be a real-world scenario where this would occur instantly to unload/load.

WS2012R2/lisa/setupscripts/CORE_reload_modules.ps1
```diff
@@ -106,8 +106,8 @@ function CheckResult()
             Start-Sleep -s 10
             $attempts--
             Write-Output "Info : Attempt number ${attempts}"
-            if ((Get-VMIntegrationService $vmName | ?{$_.name -eq "Heartbeat"}).PrimaryStatusDescription -eq "Lost Communication") {
-                Write-Output "Error : Lost Communication to VM" | Out-File -Append $summaryLog
+            if ((Get-VMIntegrationService $vmName -ComputerName $hvServer | ?{$_.name -eq "Heartbeat"}).PrimaryStatusDescription -match "No Contact|Lost Communication" ) {
+                Write-Output "Error : Lost Communication or No Contact to VM, maybe vm reboots" | Out-File -Append $summaryLog
```
chvalean: can you please also introduce here a stop-vm in force mode?
if the VM will get stuck, the lisa method of shutting down the VM might fail, as the VM is unresponsive and cannot process a graceful shutdown.

Verified it with:
stop-vm -vmName $vmName -computername $hvServer -force -turnoff

lixuemin2016:
@chvalean, understand your consideration, if add sleep 1 second after unload/load, it will sleep 6 seconds for 3 modules, it needs about 12 minutes to finish 100 rounds. If only set 100 loop count, sometimes hard to detect this bug. If set as 200 loop count, it about 24 minutes to finish test, it looks too long run time in core test. What's your suggestion about this scenario? Add new cases to stress test as your original comment?
Thank you.

WS2012R2/lisa/setupscripts/CORE_reload_modules.ps1
lixuemin2016: Correct, thank you for your comment, will update

chvalean:
Unfortunately we have to keep the sleep for a proper module reload.
I would suggest to maybe keep the SMP variation under the Core section, and under Stress add the same test but with a higher loop count and on single core.
For stress maybe it might be worth checking in 2-3 hours how many iterations can be done.
That might be sufficient as a baseline, and as you added the loop count as a param, it can be changed based on individual needs.

lixuemin2016:
@chvalean， thank you for your suggestion.
Updated based on comment, change back the default loop count as 100,  running time about 11 minutes. Also add duration time parameter, locally will set sleep time as 0 only to track regression issue in Core test.
For stress test, will add sleep 1 second, we run core test for every compose, but not stress test.

PR #1104, opened 2018-09-05
Updated INST_LIS_TestVMHeartbeat

zhanghb97:
Check the VMs heartbeat during booting up


chvalean:
Thank you for the code changes @zhanghb97 , however I think this scenario is not necessary.
There is the initial heartbeat check in the begging of the script, which implies that a heartbeat must be detected before anything else - including after the boot-up process, otherwise it will fail from the start.

If there is any background information on the heartbeat status during the boot process please let me know.

zhanghb97:
Thanks for your advice @chvalean . There is a "No Contact" status during the VM's booting up. I am wondering if it's necessary to check the "No Contact" status to make sure the heartbeat service is going well.

PR #1136, opened 2018-11-14
Rename eventid_18602_regression to MultipleReboot.ps1

lixuemin2016:
Hi Chris,
Looks that before original MultipleReboot.ps1 has been removed and replaced as eventid_18602_regression,  running list in the fvt.xml files use MultipleReboot.ps1.  So change back and refine some log output.

Thank you so much.

### DIFF
--- a/WS2012R2/lisa/setupscripts/eventid_18602_regression.ps1
+++ b/WS2012R2/lisa/setupscripts/eventid_18602_regression.ps1
@@ -20,12 +20,12 @@ function Wait-VMState {
     $currentRetryCount = 0
     while ($currentRetryCount -lt $RetryCount -and `
               (Get-VM -ComputerName $hvServer -Name $vmName).State -ne $VMState) {
-        Write-Host "Waiting for VM ${VMName} to enter ${VMState} state"
+        Write-Output "Waiting for VM ${VMName} to enter ${VMState} state"
         Start-Sleep -Seconds $RetryInterval
         $currentRetryCount++
     }
     if ($currentRetryCount -eq $RetryCount) {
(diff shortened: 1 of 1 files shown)


chvalean:
Hi @lixuemin2016 .
Based on previous test data the conclusion was that the original multiple reboot test was not catching boot failures, even if any, so we decided to replace it with a newer situation based on the mentioned commit ID.
If for some reasons you still want to use the previous MultipleReboot test, can you please make that as a separate test case in the xml?
That way you can still use it as fit, the code was already there.

Thank you!

lixuemin2016:
Hi Chris, yes, file eventid_18602_regression is very good to detect reboot failure, I do not want to use previous MultipleReboot test. 

Two reasons about this change:
1) Original MultipleReboot.py has been removed based on PR https://github.com/LIS/lis-test/commit/fdc9b655cdfcfab8991c8d41056c4f039d8833ed#diff-7cc5af26bdc6fe093b7bf5ad19c10634, notice that hyperv_fvt.xml, kernel_pipeline_fvt.xml,lis_pipeline_fvt,ubuntu_azure_kernel_fvt.xml still use MultipleReboot.py but not eventid_18602_regression.py except EventID_18602_regression test case in StressTests.xml.
2) eventid_18602_regression is mainly  used to test reboot vm, maybe file name "MultipleReboot.py" is more easy to indicate the main test points? 
Thank you so much.

chvalean:
Please keep the code changes in the script, but leave the naming as is.
The referenced xml files do indeed carry the old test definition, however we actually do not run it - it's not in the list of tests, mainly due to the long time required for the test, so we keep it separate in the stress xml.

lixuemin2016:
Done. Thank you.

PR #1139, opened 2018-11-19
Update kdump for grub2-editenv to set kernelopts

lixuemin2016:
Hi Chris,
Refer to bug https://bugzilla.redhat.com/show_bug.cgi?id=1637875, GRUB_CMDLINE_LINUX does not work probably for some reason. From the comment of developer, if kernelopts and other grub.cfg variables are defined in /boot/efi/EFI/redhat/grubenv for EFI systems and /boot/grub2/grubenv for legacy BIOS. This file should not be edited with a text editor. It can be modified using:
grub2-editenv - set kernelopts=”rd.debug=1 rhgb”.
Based on above reason, do minor change for Kdump_Config.sh

Thank you.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/Kdump_Config.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/Kdump_Config.sh
@@ -424,14 +424,11 @@ case $DISTRO in
     ;;
 esac
 
-grep -i vmlinuz $boot_filepath
-if [ $? -ne 0 ]; then
-    # Can not find kernel command in grub.cfg, need to update /etc/default/grub
-    default_grub="/etc/default/grub"
-    sed -i "s/crashkernel=\S*//g" $default_grub
-    sed -i "s/console=\S*//g" $default_grub
-    sed -i "/GRUB_CMDLINE_LINUX/ s/\"$/ crashkernel=$crashkernel\"/" $default_grub
-    grub2-mkconfig -o $boot_filepath
+command -v grub2-editenv
+if [ $? -eq 0 ] && [[ ! -z `grub2-editenv - list | grep -i kernelopts` ]]; then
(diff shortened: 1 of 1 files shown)


WS2012R2/lisa/remote-scripts/ica/Kdump_Config.sh
```diff
@@ -424,14 +424,11 @@ case $DISTRO in
...
-grep -i vmlinuz $boot_filepath
-if [ $? -ne 0 ]; then
-    # Can not find kernel command in grub.cfg, need to update /etc/default/grub
-    default_grub="/etc/default/grub"
-    sed -i "s/crashkernel=\S*//g" $default_grub
-    sed -i "s/console=\S*//g" $default_grub
-    sed -i "/GRUB_CMDLINE_LINUX/ s/\"$/ crashkernel=$crashkernel\"/" $default_grub
-    grub2-mkconfig -o $boot_filepath
+# check if kernelopts available in grub2-editenv list, if yes, set kernelopts
+grub2-editenv - list | grep -i kernelopts
```
chvalean: I think it might be worth checking first if grub2-editenv command is present, and only at the end fallback to the sed option.

lixuemin2016: Chris, thank you for your review and comment.
The command "grub2-editenv" is also present in rhel7 and fedora.
e.g. on Fedora,
"grub2-editenv -list" output saved_entry=Fedora (4.17.12-100.fc27.x86_64) 27 (Workstation Edition)
so think about using "grub2-editenv - list | grep -i kernelopts" to distinguish different methods or releases to replace directly check distribution. Is it reasonable? Thank you so much.

lixuemin2016:
Understand, will update. Thank you.

PR #1140, opened 2018-11-20
Refine SR-IOV_Ping_DisableVF one value compare to average value

lixuemin2016:
Before SR-IOV_Ping_DisableVF compare one value of PingResults.log, which makes test case not stable, Sometimes test case failed, after rerun maybe pass. Comparing RTT average value(enable, disable, re-enable again) during 30 seconds is more reliable than one instant value.

Thank you.


chvalean:
@adriansuhov please check, this is interesting.
Thank you @lixuemin2016 for the contribution!

PR #1142, opened 2018-11-21
kdump_nfs: check modules for kdump.img before execute kdump, skip che…

HuijingHei:
1. Add modules check based on bug https://bugzilla.redhat.com/show_bug.cgi?id=1603017
2. Skip modules check when kernel version older than kernel-3.10.0-957.el7

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/modules_check.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/modules_check.sh
@@ -155,20 +155,31 @@ case $DISTRO in
     ;;
 esac
 
-if [ -f /boot/initramfs-0-rescue* ]; then
-    img=/boot/initramfs-0-rescue*
+if [ "${img:-UNDEFINED}" = "UNDEFINED" ]; then
+    if [ -f /boot/initramfs-0-rescue* ]; then
+        img=/boot/initramfs-0-rescue*
+    else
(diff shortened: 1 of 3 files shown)


WS2012R2/lisa/remote-scripts/ica/modules_check.sh
```diff
@@ -155,20 +155,31 @@ case $DISTRO in
...
 
-  if [ -f "/boot/initrd.img-`uname -r`" ]; then
-    img="/boot/initrd.img-`uname -r`"
-  fi
+    if [ -f "/boot/initrd.img-`uname -r`" ]; then
+        img="/boot/initrd.img-`uname -r`"
+    fi
+    fi
+else
+    if [ $img == "kdump.img" ] && [ -f /boot/initramfs-`uname -r`kdump.img ]; then
```
chvalean: I don't think we need this - $img == "kdump.img"
The if for /boot/initramfs-`uname -r`kdump.img should go in line with the previous checks for different image names.
Please also remove then the new variable from xml

HuijingHei: Hi Chris, thanks for your review.  I add the $img variable to check the modules whether included in "initramfs-uname -rkdump.img", without it, will check /boot/initramfs-0-rescue*, so $img makes sense. If I am wrong, you can correct me, thanks!

PR #1145, opened 2018-12-11
Minor update DM_HotAdd_Stress_ng.ps1 to add stressLevel 3 which is lo…

lixuemin2016:
…wer stress

Add lower stress_Level 3 by using half of threads which is used to reach the maximum memory for stress-ng and small chunks size 512. If use stress_Level 0,1,2 makes VM out of memory sometimes, it will trigger call trace, the test case will fail. 
To make test case more stable, use lower stress. If it gets call trace, means exact bug but not triggered by stress-ng.

Thank you.

### DIFF
--- a/WS2012R2/lisa/setupscripts/DM_HotAdd_Stress_ng.ps1
+++ b/WS2012R2/lisa/setupscripts/DM_HotAdd_Stress_ng.ps1
@@ -239,6 +239,11 @@ elseif ($timeoutStress -eq 2) {
     $duration = 40
     $chunk = 1
 }
+elseif ($timeoutStress -eq 3) {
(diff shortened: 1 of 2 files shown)


lixuemin2016:
@chvalean 
Hi Chris, I just find HotAddStressapptest also can reach the lower stress, if you think current stressLevel 3 for stress-ng is not good PR, I will close it.
Thank you.

lixuemin2016:
HotAddStressapptest can be used to do lower stress

chvalean:
Thank you @lixuemin2016 for the comments.
Memory hot add operations in a stress pattern is usually difficult to predict to behave a certain way from my experience, so I would agree with you.

PR #1146, opened 2018-12-19
Add one balloon case to check assigned memory goes down when no pressure

lixuemin2016:
Add one case HotRemoveNoPressure for balloon feature, it mainly tests that  vm assigned memory could go down after sleep about 4-7 minutes without pressure, but not start the other guestB with large memory to make guestA Balloon (assigned memory goes down)
Test case steps:
1. Config VM as Dynamic memory enabled.
2. Set up the VM  Startup Memory, MinMem,MaxMem. (set Startup Memory as larger size)
3. Start VM , sleep about 4-7 minutes till the assigned and demand memory go down.

Expected results:
3.1. VM  assigned memory and demand memory can go down without memory pressure.
3.2. VM assigned memory does not drop lower then minimum memory
3.3. No call trace in the vm.

Thank you so much.

### DIFF
--- a/WS2012R2/lisa/setupscripts/DM_RemoveNoPressure.ps1
+++ b/WS2012R2/lisa/setupscripts/DM_RemoveNoPressure.ps1
@@ -1,0 +1,177 @@
+#####################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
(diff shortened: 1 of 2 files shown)


lixuemin2016:
Will add one more case which could cover this test case, remove this pull request.

PR #1147, opened 2018-12-20
Add balloon case

lixuemin2016:
Following two changes for this PR:

Change #1:
Add one new case HotRemoveThenPressure, currently existing balloon test cases do not cover this scenario. When test with two vms,  it will use most of host memory, test cases sometime become not stable. This case test on one vm.
     Step 1: Verify that a VM's assigned memory could decrease when no pressure available.
     After VM sleeps less than 7 minutes, it is higher than minimum memory.
     Step 2: Do stress-ng test, during stress test, assigned and demand memory increase
     Step 3: After stress test, check that assigned and memory decrease again, no any crashes.

Note: Here do not use the ConsumeMemory in the TCUtils.ps1, because local function executes the stress based on free memory, then it has lower level stress and does not trigger the out of memory call trace.

Change #2:
DM_stressng -> sometimes trigger call trace, ooo . (Add maxMemratio to adapt stress.)

Thank you so much.


chvalean:
@lixuemin2016  for the new testcase, I'm being told that it fails when using Min memory 1024MB, but it does pass if changing that to 512MB.
Can you please advise on what release/kernel has this been tested or expected to work?

@asofron for ack

PR #1154, opened 2019-02-27
Add new case about verifying CPU can be offline permanently

xernyuii:
This case is about verifying CPU can be offline permanently by boot with nr_cpus=xx. (Bug 1396336, Bug 1396335)

Test step:
1. Set the vcpu number.
2. Add nr_cpus=xx in kernel parameter and reboot it.
3. Verify the exact CPUs number and try to offline the CPUs.

Verify results:
1. Exact CPUs number and nr_cpus should be equal.
2. CPUs with VMBus channels attached should not go offline.

Thank you.

### DIFF
--- a/WS2012R2/lisa/remote-scripts/ica/Core_Offline_CPU_SMP.sh
+++ b/WS2012R2/lisa/remote-scripts/ica/Core_Offline_CPU_SMP.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+
+########################################################################
+#
+# Linux on Hyper-V and Azure Test Code, ver. 1.0.0
+# Copyright (c) Microsoft Corporation
+#
+# All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the ""License"");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


chvalean:
Thank you for the submission @xernyuii .
Can you please check the existing test case VCPU_verify_online - it does the same check, without the grub parameter.
Otherwise, let's discuss how we can merge these 2 test cases.

xernyuii:
Thanks for your suggestion. @chvalean 
I read the test case VCPU_verify_online. May be I should just verify offline CPUs before modifying grub parameter? And it will cover that case (replace the  VCPU_verify_online).

HuijingHei:
@xernyuii , thanks for your effort. I clarify it as this:
VCPU_verify_online - check the vcpus should be online with the setting
Core_Offline_CPU_SMP - check the vcpus should be online with nr_cpus=N in kernel parameter 

If check vcpus should be online before add nr_cpus=N in kernel parameter in the Core_Offline_CPU_SMP, then it can cover VCPU_verify_online.

@chvalean, What's your suggestion? Thanks!

